PR #2, opened 2021-03-10
Set up unicorn

barrucadu:
Our deployment and monitoring scripts assume that we use unicorn, not
puma.  This is a shame, as the default Rails server has been puma for
years, but fixing this would be a nontrivial change to a
rarely-touched part of the GOV.UK stack.

---

[Trello card](https://trello.com/c/ZeAKl2SM/651-set-up-a-new-app)


huwd (approved):

AlanGabbianelli (approved):
😢

PR #3, opened 2021-03-10
Implement authentication endpoints

barrucadu:
This PR adds the three authentication endpoints from the RFC, at 95.41% line coverage (enough for continuous deployment).

The way the auth flow will work for the Transition Checker is:

1. finder-frontend calls `/api/oauth2/state` with the attributes, which returns a `state_id`
2. finder-frontend redirects the user to `/sign-in?state_id={...}&redirect_path={...}` (which is rendered by frontend)
3. frontend calls `/api/oauth2/sign-in` with the `state_id` and `redirect_path`, which returns an `auth_uri` and a `state`
4. frontend stores the `state` in the session (this is for CSRF protection) and redirects the user to the `auth_uri`
5. the user authenticates, and is sent to `/sign-in/callback?code={...}&state={...}` (which is rendered by frontend)
6. frontend checks that the `state` param matches the value in the session (for CSRF protection), and calls `/api/oauth2/callback` with the `code` and the `state`, which returns a `govuk_account_session`, `redirect_path`, and `ga_client_id`
7. frontend stores the `govuk_account_session` in the session, and redirects to `{redirect_path}?_ga={ga_client_id}`
8. the user ends up back in the transition checker, logged in

Much of the logic is copied and pasted from finder-frontend.

---

[Trello card](https://trello.com/c/Y8z5rsJl/653-move-oauth-sign-in-logic-from-the-transition-checker-to-the-new-app)

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -1,5 +1,7 @@
 class AuthenticationController < ApplicationController
   def sign_in
+    AuthRequest.expired.delete_all
+
     auth_request = AuthRequest.create!(
       oauth_state: params.fetch(:state_id, SecureRandom.hex(16)),
       oidc_nonce: SecureRandom.hex(16),
@@ -25,7 +27,7 @@ class AuthenticationController < ApplicationController
 
     redirect_path = auth_request.redirect_path
 
-    auth_request.delete!
+    auth_request.delete
 
     render json: {
       govuk_account_session: to_account_session(oauth_response[:access_token], oauth_response[:refresh_token]),
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -1,7 +1,40 @@
 class AuthenticationController < ApplicationController
-  def sign_in; end
+  def sign_in
+    auth_request = AuthRequest.create!(
+      oauth_state: params.fetch(:state_id, SecureRandom.hex(16)),
+      oidc_nonce: SecureRandom.hex(16),
+      redirect_path: params[:redirect_path],
+    )
 
-  def callback; end
+    render json: {
+      auth_uri: OidcClient.new.auth_uri(auth_request),
+      state: auth_request.to_oauth_state,
+    }
+  end
+
+  def callback
+    auth_request = AuthRequest.from_oauth_state(params.fetch(:state))
+    head :unauthorized and return unless auth_request
+
+    client = OidcClient.new
+    tokens = client.callback(auth_request, params.fetch(:code))
+    oauth_response = client.get_ephemeral_state(
+      access_token: tokens[:access_token],
+      refresh_token: tokens[:refresh_token],
+    )
+
+    redirect_path = auth_request.redirect_path
+
+    auth_request.delete!
+
+    render json: {
+      govuk_account_session: to_account_session(oauth_response[:access_token], oauth_response[:refresh_token]),
+      redirect_path: redirect_path,
+      ga_client_id: oauth_response[:result]["_ga"],
+    }
+  rescue OidcClient::OAuthFailure
+    head :unauthorized
+  end
 
   def create_state; end
 end
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -36,5 +36,21 @@ class AuthenticationController < ApplicationController
     head :unauthorized
   end
 
-  def create_state; end
+  def create_state
+    payload = {
+      attributes: params[:attributes].permit!.to_h,
+    }.compact
+
+    client = OidcClient.new
+    tokens = client.tokens!
+    oauth_response = client.submit_jwt(
+      jwt: JWT.encode(payload, nil, "none"),
+      access_token: tokens[:access_token],
+      refresh_token: tokens[:refresh_token],
+    )
+
+    render json: {
+      state_id: oauth_response[:result]["id"],
+    }
+  end
 end
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -1,0 +1,7 @@
+class AuthenticationController < ApplicationController
+  def sign_in; end
+
+  def callback; end
+
+  def create_state; end
+end
--- a/app/lib/oidc_client.rb
+++ b/app/lib/oidc_client.rb
@@ -1,0 +1,129 @@
+require "openid_connect"
+
+class OidcClient
+  class OAuthFailure < RuntimeError; end
+
+  DEFAULT_SCOPES = %i[transition_checker openid].freeze
+
+  attr_reader :client_id,
+              :provider_uri
+
+  delegate :authorization_endpoint,
+           :token_endpoint,
+           :userinfo_endpoint,
+           :end_session_endpoint,
+           to: :discover
+
+  def initialize(provider_uri: nil, client_id: nil, secret: nil)
+    @provider_uri = provider_uri || Plek.find("account-manager")
+    @client_id = client_id || ENV.fetch("GOVUK_ACCOUNT_OAUTH_CLIENT_ID")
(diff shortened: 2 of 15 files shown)


app/controllers/authentication_controller.rb
```diff
@@ -1,7 +1,40 @@
...
+    )
+
+    redirect_path = auth_request.redirect_path
+
+    auth_request.delete!
+
+    render json: {
+      govuk_account_session: to_account_session(oauth_response[:access_token], oauth_response[:refresh_token]),
+      redirect_path: redirect_path,
+      ga_client_id: oauth_response[:result]["_ga"],
```
huwd, line 35: This looks good but seems far less involved than some of the things we had to do in the pervious app to preserve `_ga` around these various stages.

Is that just because we're on the same domain now?

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,129 @@
...
+    begin
+      response.merge(result: JSON.parse(response[:result].body))
+    rescue JSON::ParserError
+      response.merge(result: {})
+    end
+  end
+
+protected
+
+  OK_STATUSES = [200, 204, 404, 410].freeze
```
huwd, line 89: Why are 404, 410 OK states?

app/lib/oidc_client.rb
```diff
@@ -0,0 +1,129 @@
...
+      access_token: access_token_str,
+      refresh_token: refresh_token_str,
+      result: response,
+    }
+  rescue AttrRequired::AttrMissing, Rack::OAuth2::Client::Error, URI::InvalidURIError
+    raise OAuthFailure
+  end
+
+  def redirect_uri
+    host = Rails.env.production? ? ENV["GOVUK_WEBSITE_ROOT"] : Plek.find("finder-frontend")
```
huwd, line 122: Could you talk me through the thinking here too?
So if we're prod we're redirecting to root, if not we're going to finder-frontend?

I assume that's to preserve our dev setup right now on subdomains?
But wouldn't that also break integration / staging?

That'll need to be changed when we start adding 2nd services?

spec/requests/authentication_controller_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    discovery_response = instance_double(
+      "OpenIDConnect::Discovery::Provider::Config::Response",
+      authorization_endpoint: "http://openid-provider/authorization-endpoint",
+      token_endpoint: "http://openid-provider/token-endpoint",
+      userinfo_endpoint: "http://openid-provider/userinfo-endpoint",
+      end_session_endpoint: "http://openid-provider/end-session-endpoint",
+    )
+
+    # rubocop:disable RSpec/AnyInstance
+    allow_any_instance_of(OidcClient).to receive(:discover).and_return(discovery_response)
```
huwd: How come we need `allow_any_instance_of` here?
Is this to save us having to put an `allow()` in every block where we're defining an instance of this class with a `let`?

spec/lib/oidc_client_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  describe "tokens!" do
+    before do
+      access_token = Rack::OAuth2::AccessToken.new(
+        access_token: "access-token",
+        refresh_token: "refresh-token",
+        token_type: "test",
+      )
+
+      @client_stub = stub_oidc_client(client)
+      # rubocop:disable RSpec/InstanceVariable
```
huwd, line 15: Oh, humm, rubocop doesn't like instance allows either?
This seems like a reasonable syntax to me... I wonder why we put this rule on 🤔

spec/requests/authentication_controller_spec.rb
```diff
@@ -31,6 +31,12 @@
       get sign_in_path(redirect_path: "/hello-world")
       expect(AuthRequest.last.redirect_path).to eq("/hello-world")
     end
+
+    it "deletes old expired AuthRequests" do
```
huwd, line 35: Are we going to have any kind of job for this too?
Like the thing we used to periodically clean up the RegistrationState and LoginState?

huwd: Ah ha, just seen the commit message,
Ok sure.

I guess only thought here is, is there any peril in us having users who don't log in for a while who have crufy state hanging around?

Won't we need to set up sidekiq for other things too?

If not, then I think I agree this is a nice and simple approach.

app/models/auth_request.rb
```diff
@@ -14,4 +16,19 @@ def self.from_oauth_state(state)
...
+    return if redirect_path.nil?
+    return if redirect_path.empty?
+
+    if redirect_path&.starts_with? "//"
+      errors.add(:redirect_path, "can't be protocol-relative")
+      return
+    end
+
+    return if redirect_path.starts_with? "/"
+    return if redirect_path.starts_with?("http://") && Rails.env.development?
```
huwd, line 30: ... I still think we should make dev work with https... but great thinking ahead fr the current setup!

spec/models/auth_request_spec.rb
```diff
@@ -14,4 +14,22 @@
     expect(described_class.from_oauth_state(request1.to_oauth_state).id).to eq(request1.id)
     expect(described_class.from_oauth_state(request2.to_oauth_state).id).to eq(request2.id)
   end
+
```
huwd, line 17: Nice set of tests

huwd (approved):
A few things I'd like to discuss / understand better but the code looks clean, familliar in the right places and makes an overall sense.

Nice one!

app/controllers/authentication_controller.rb
barrucadu: It's simple here because this isn't user facing, it's just passing parameters around.

In finder-frontend we still have the `redirect_with_ga` thing: https://github.com/alphagov/finder-frontend/pull/2457/files#diff-cdec550eeeb8fc63be2b5687170f594651a8d0b7f0465c9d807baef392639b6eR26

app/lib/oidc_client.rb
barrucadu: They can be legitimate API responses, we only want the refresh-tokens-and-retry logic to kick in if the tokens are rejected or there's an unexpected response.

app/lib/oidc_client.rb
barrucadu: > I assume that's to preserve our dev setup right now on subdomains?

Yeah.  When we do https://trello.com/c/RaI017p6/730-add-new-public-facing-oauth-endpoints-to-frontend, it'll change to `Plek.find("frontend")`

> But wouldn't that also break integration / staging?

No, Rails environments are `development`, `test`, and `production` - which correspond to running the app locally, running the tests, or running it for real.  Our integration, staging, and production environments are all "production" asfar as Rails is concerned.

spec/requests/authentication_controller_spec.rb
barrucadu: The tests don't have access to the actual `OidcClient` instance because they're testing via the controllers.  An alternative way of doing this would be to use webmock to stub the discovery response, but that's kind of awkward because the `openid_connect` gem expects everything to be set up with proper crypto.

spec/requests/authentication_controller_spec.rb
barrucadu: > I guess only thought here is, is there any peril in us having users who don't log in for a while who have crufy state hanging around?

That'll only happen if we have a bunch of auth requests created in a short period of time, and then no more for hours.  So I don't think it's much of a risk.

> Won't we need to set up sidekiq for other things too?

We'll probably end up needing to export attributes from here to bigquery, or something, which would be done best with sidekiq.  But we don't need it now and after the pain of getting the app running I'm not super keen on also getting sidekiq running...

app/lib/oidc_client.rb
huwd: Ace, that makes sense now thanks very much

PR #5, opened 2021-03-11
Add endpoints to manage the Transition Checker email subscription 

barrucadu:
By putting these endpoints in this app, we can fully remove the OIDC /
OAuth logic from the Transition Checker.

The name of the controller is really specific because we've not yet
got a generic way to manage email subscriptions for multiple apps, and
will need to think about how to do that if it's something we want.

---

[Trello card](https://trello.com/c/N49KKdOW/666-move-transition-checker-email-management-from-finder-frontend-to-the-new-app)

### DIFF
--- a/app/controllers/transition_checker_email_subscription_controller.rb
+++ b/app/controllers/transition_checker_email_subscription_controller.rb
@@ -1,0 +1,31 @@
+class TransitionCheckerEmailSubscriptionController < ApplicationController
+  before_action :fetch_govuk_account_session
+
+  def show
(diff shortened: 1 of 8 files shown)


app/controllers/transition_checker_email_subscription_controller.rb
```diff
@@ -0,0 +1,31 @@
+class TransitionCheckerEmailSubscriptionController < ApplicationController
```
huwd, line 1: A long specific name seems better than a short ambigious one!

huwd (approved):
Looks great once again!

PR #7, opened 2021-03-11
Add a Jenkinsfile

huwd:
Following these instructions here: https://docs.publishing.service.gov.uk/manual/setting-up-new-rails-app.html#configuring-the-app-for-jenkins

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,7 @@
+#!/usr/bin/env groovy
+
+library("govuk@default-branch")
+
+node {
+  govuk.buildProject()
+}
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env groovy
+
+library("govuk")
```
barrucadu, line 3: govuk-jenkinslib assumes the default branch is `master`, use this branch of it instead:

```
library("govuk@default-branch")
```

huwd: ace thanks very much @barrucadu !

barrucadu (approved):

PR #8, opened 2021-03-12
Add endpoints to manage attributes 

barrucadu:
Currently these just delegate to the govuk-attribute-service-prototype
running on the PaaS, but at some point soon we will begin storing
attributes in this app's database too.  We might migrate the
Transition Checker attributes, or we might just retire the checker -
that depends on how long it'll be staying around, which is a product
decision.

For the update endpoint, the attribute values are given as
JSON-encoded strings.  I think this is reasonable, because there isn't
really a standard for how complex nested data gets turned into HTTP
field data.  We can do the `to_json`ing in gds-api-adapters, so that
other apps don't need to worry about it.

---

[Trello card](https://trello.com/c/uRMyiEHZ/654-move-oauth-attribute-logic-from-the-transition-checker-to-the-new-app)

### DIFF
--- a/app/controllers/attributes_controller.rb
+++ b/app/controllers/attributes_controller.rb
@@ -1,0 +1,44 @@
+class AttributesController < ApplicationController
+  before_action :fetch_govuk_account_session
+
+  def show
+    attribute_names = params.fetch(:attributes)
+
+    access_token = @govuk_account_session[:access_token]
+    refresh_token = @govuk_account_session[:refresh_token]
+
+    values = attribute_names.each_with_object({}) do |name, values_hash|
+      oauth_response = OidcClient.new.get_attribute(
+        attribute: name,
+        access_token: access_token,
+        refresh_token: refresh_token,
+      )
+      access_token = oauth_response[:access_token]
+      refresh_token = oauth_response[:refresh_token]
+      values_hash[name] = oauth_response[:result]
+    end
+
+    render json: {
+      govuk_account_session: to_account_session(@govuk_account_session[:access_token], @govuk_account_session[:refresh_token]),
+      values: values.compact,
+    }
+  rescue OidcClient::OAuthFailure
+    head :unauthorized
+  end
+
+  def update
+    attributes = params.fetch(:attributes).permit!.to_h
+
+    oauth_response = OidcClient.new.bulk_set_attributes(
+      attributes: attributes,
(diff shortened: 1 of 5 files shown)


app/controllers/attributes_controller.rb
```diff
@@ -0,0 +1,44 @@
+class AttributesController < ApplicationController
+  before_action :fetch_govuk_account_session
+
+  def show
+    attribute_names = params.fetch(:attributes)
+
+    access_token = @govuk_account_session[:access_token]
+    refresh_token = @govuk_account_session[:refresh_token]
+
+    values = attribute_names.each_with_object({}) do |name, hsh|
```
huwd, line 10: nitpick, I see you assign it to the `values` variable here.
but not sure how I feel about `hsh` as shorthand. 

Could we shadow this as `values` here in the iterator, or at least use `hash`?

Total nitpick, take it or leave it.

app/controllers/attributes_controller.rb
```diff
@@ -0,0 +1,44 @@
+class AttributesController < ApplicationController
```
huwd, line 1: Just checking we're doing a lot of pulling keys out of hashes in this controller.

They all seem to be coming from params or http-recieved objects that rails has passed.

The keys here are all symbols. I remember this catching us out in the past, no?

Are we sure that JSON processed string key names get processed into symbols?

spec/requests/attributes_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
+RSpec.describe AttributesController do
+  before { stub_oidc_discovery }
+
+  let(:headers) { { "GOVUK-Account-Session" => placeholder_govuk_account_session } }
+
+  let(:attribute_name1) { "name" }
+  let(:attribute_value1) { { "some" => "complex", "value" => 42 } }
```
huwd, line 7: This example would seem to suggest the string -> sym transition works pretty well

huwd (approved):
One thing to check and a super ignorable nitpick.
But looks good, nice clean methods!

app/controllers/attributes_controller.rb
barrucadu: > Are we sure that JSON processed string key names get processed into symbols?

Actually they're all strings, no symbols involved at all.

PR #13, opened 2021-03-17
Write Pact provider configuration

barrucadu:
This is step 2 of the [Writing Pact tests](https://docs.publishing.service.gov.uk/manual/pact-broker.html#writing-pact-tests) docs, see also https://github.com/alphagov/gds-api-adapters/pull/1043.

---

Pact provider configuration specifies which contracts are
honoured (currently only gds-api-adapters) and defines a set of
"provider states", which are a sort of precondition that the consumer
tests expect.

A Pact test has two parts:

1. The consumer test, which specifies which provider state is
expected, gives the request, and an expected response.

2. The provider configuration, which defines all of the provider
states.

There are no actual tests in the provider: rather, the requests
defined by the consumer tests are run against the provider, and pact
checks that the responses match.

---

[Trello card](https://trello.com/c/OcRffYwv/674-set-up-pact-tests-for-our-account-api-app)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,8 @@ group :development, :test do
   gem "bullet"
   gem "factory_bot_rails"
   gem "govuk_test"
+  gem "pact", require: false
+  gem "pact_broker-client"
   gem "pry-byebug"
   gem "pry-rails"
   gem "rspec-rails"
--- a/app/models/auth_request.rb
+++ b/app/models/auth_request.rb
@@ -4,6 +4,14 @@ class AuthRequest < ApplicationRecord
 
   validate :redirect_path_is_safe
 
+  def self.generate!(options = {})
+    create!(
+      oauth_state: options[:oauth_state] || SecureRandom.hex(16),
+      oidc_nonce: SecureRandom.hex(16),
+      redirect_path: options[:redirect_path],
+    )
+  end
+
   # This has to be something which the account manager can use to retrieve a JWT, if there is one:
   # https://github.com/alphagov/govuk-account-manager-prototype/blob/6a68e02055fe3c70083b3899b474b10cc944ffa5/config/initializers/doorkeeper.rb#L14
   def to_oauth_state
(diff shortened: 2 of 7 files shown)


Gemfile
```diff
@@ -18,6 +18,8 @@ group :development, :test do
   gem "bullet"
   gem "factory_bot_rails"
   gem "govuk_test"
+  gem "pact", require: false
```
ChrisBAshton, line 21: Typo in commit message (`anc` => `and`)

spec/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,141 @@
...
+  config.include WebMock::API
+  config.include WebMock::Matchers
+end
+
+Pact.service_provider "Account API" do
+  honours_pact_with "GDS API Adapters" do
+    if ENV["PACT_URI"]
+      pact_uri ENV["PACT_URI"]
+    else
+      abort("Not set up to work with the pact-broker yet, only local usage is implemented.
```
ChrisBAshton, line 15: Nice 👍

ChrisBAshton (commented):
Approach LGTM 👍

spec/service_consumers/pact_helper.rb
hannako: This is cool

app/models/auth_request.rb
```diff
@@ -4,6 +4,14 @@ class AuthRequest < ApplicationRecord
 
   validate :redirect_path_is_safe
 
+  def self.generate!(options = {})
```
huwd, line 7: I see, so you've pushed the logic down a bit, so we do all the defaults / logic here?

Was this a general tidy up thing you noticed whilst writing these, or were the Pact tests giving you grief with the old `sign_in` method?

huwd (approved):

app/models/auth_request.rb
barrucadu: Just a general tidy up.  I'd ended up with three copies of the code to create an `AuthRequest`, so I decided to put it in one place.

PR #16, opened 2021-03-18
Send application/json requests in the request specs

barrucadu:
By default, Rails request specs send parameters as
application/x-www-form-urlencoded, but this has two problems:

1. gds-api-adapters sends JSON requests, so sending form data means
our test & production behaviours are very different.

2. form-data is bad at encoding complex nested data (or even just
numbers - which are decoded as strings), so we limit our app to only
what form-data supports.

A concrete problem caused by (2) is that the PATCH attributes endpoint
requires attribute values to be passed in as strings of JSON-encoded
data.  This is weird and confusing, and not at all like publishing-api
or email-alert-api.  It would be good to fix that, but first we need
these tests to do the right thing.

By instead sending JSON data in our tests (which publishing-api does
in its tests), we make our tests more closely match production
behaviour.

---

[Trello card](https://trello.com/c/E5J5kzNP/679-remove-the-redundant-json-encoding-of-attribute-values-in-account-api)

### DIFF
--- a/spec/requests/authentication_controller_spec.rb
+++ b/spec/requests/authentication_controller_spec.rb
@@ -7,6 +7,8 @@ RSpec.describe AuthenticationController do
     # rubocop:enable RSpec/AnyInstance
   end
 
+  let(:headers) { { "Content-Type" => "application/json" } }
+
   describe "/sign-in" do
     it "creates an AuthRequest to persist the attributes" do
       expect { get sign_in_path }.to change(AuthRequest, :count)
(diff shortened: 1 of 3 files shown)


spec/requests/authentication_controller_spec.rb
```diff
@@ -23,12 +25,12 @@
     end
 
     it "uses a provided state_id" do
-      get sign_in_path(state_id: "hello-world")
+      get sign_in_path, params: { state_id: "hello-world" }
```
huwd, line 28: This and L33 stand out as the only two param blocks not getting the `.to_json` treatment here.
So is the rule that we _only_ want to transform to JSON for things that are nested?

Otherwise they stay normal params?

barrucadu: These are GET requests, so the params are sent in the querystring, not in the body.

huwd (approved):

PR #17, opened 2021-03-18
Allow non-encoded attribute values

barrucadu:
The create-state endpoint accepts plain JSON, rather than encoded
values.  This commit makes the update-attributes endpoint also accept
plain JSON, to be consistent.

To support backwards compatibility, encoded attribute values are also
accepted.  If all of the attribute values are JSON strings, then the
assumption is made that they have been encoded by the sender.  In
practice, this is the case, because gds-api-adapters is the only thing
which uses this endpoint so far and that's what it does.

After this has been deployed we can:

1. update gds-api-adapters to *not* encode the attribute values
2. deploy a finder-frontend with the updated gds-api-adapters
3. remove the backwards compatibility code here

Moving to accept plain JSON values (rather than strings of encoded
values) makes this endpoint consistent with other GOV.UK API apps.

---

[Trello card](https://trello.com/c/E5J5kzNP/679-remove-the-redundant-json-encoding-of-attribute-values-in-account-api)

### DIFF
--- a/app/controllers/attributes_controller.rb
+++ b/app/controllers/attributes_controller.rb
@@ -29,8 +29,11 @@ class AttributesController < ApplicationController
   def update
     attributes = params.fetch(:attributes).permit!.to_h
 
+    already_encoded = attributes.values.all? { |v| is_json_encoded v }
+
     oauth_response = OidcClient.new.bulk_set_attributes(
       attributes: attributes,
+      already_encoded: already_encoded,
       access_token: @govuk_account_session[:access_token],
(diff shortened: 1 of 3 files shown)


app/controllers/attributes_controller.rb
```diff
@@ -41,4 +44,15 @@ def update
   rescue OidcClient::OAuthFailure
     head :unauthorized
   end
+
+protected
+
+  def is_json_encoded(value)
```
huwd, line 50: nice

huwd (approved):

PR #19, opened 2021-03-22
Flesh out README & add API docs

barrucadu:
The README is shorter than our Account Manager or Attribute Service apps, eg it doesn't have detailed set-up documentation.

Our GOV.UK READMEs have been standardising on a fairly terse format[(1)][][(2)][][(3)][] which I've copied.  In particular, there are no specific GOV.UK Docker commands referenced, I asked why on Slack and the thinking is that it makes GOV.UK Docker seem more magical than it is, and makes people think that they have to use GOV.UK Docker to run our apps, when really it's nothing more than a convenient shortcut.

The API documentation is inspired by [publishing-api][] and [router-api][].

[(1)]: https://github.com/alphagov/imminence
[(2)]: https://github.com/alphagov/router-api
[(3)]: https://github.com/alphagov/search-api
[publishing-api]: https://github.com/alphagov/publishing-api/blob/master/docs/api.md
[router-api]: https://github.com/alphagov/router-api/blob/master/docs/api-endpoints.md

---

[Trello card](https://trello.com/c/MBJ9DNmU/678-write-some-documentation-for-the-account-api)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,85 @@
 # Account API
 
-<!-- todo: Write the README -->
+Provides sign in / sign out and attribute storage functionality to
+other GOV.UK applications.
+
+In production this currently uses the [GOV.UK account manager
+prototype][] as an authentication service and the [GOV.UK attribute
+service prototype][] as a user data store.
+
+This app does not serve any user-facing pages.  To see it working you
+will need to run an app which uses it, such as [Finder Frontend][].
+
+After starting Finder Frontend, you should be able to access the
+following links:
+
+- the [Brexit checker journey start page][tc-start]
+- the [Brexit checker results page][tc-results] that reflects the
+  answers you give during the Brexit checker journey
+- the [account sign up page][tc-save-results] to save your answers
+
+When you have set up your local account, you can [sign into your
+account][account-manager].
+
+[GOV.UK account manager prototype]: https://github.com/alphagov/govuk-account-manager-prototype
+[GOV.UK attribute service prototype]: https://github.com/alphagov/govuk-attribute-service-prototype
+[Finder Frontend]: https://github.com/alphagov/finder-frontend
+[account-manager]: http://www.login.service.dev.gov.uk/
+[tc-results]: http://finder-frontend.dev.gov.uk/transition-check/results?c[]=living-ie
+[tc-save-results]: http://finder-frontend.dev.gov.uk/transition-check/save-your-results?c%5B%5D=living-ie
+[tc-start]: http://finder-frontend.dev.gov.uk/transition-check/questions
+
+
+## Technical documentation
+
+This is a Ruby on Rails app, and should follow [our Rails app
+conventions][].
+
+[our Rails app conventions]: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html
+
+**Use GOV.UK Docker to run any commands that follow.**
+
+### Testing
+
+This repository follows the standards for testing described in the
+[GOV.UK RFC on continuous deployment][]:
+
+- code coverage in excess of 95%
+- API contract tests ("pact tests") between the Account API and its
+  consumers
+- a [smoke test][] to check the application is running after a
+  deployment
+
+The default `rake` task runs all the tests and records code coverage:
+
+```sh
+bundle exec rake
+```
+
+[GOV.UK RFC on continuous deployment]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md
+[smoke test]: https://github.com/alphagov/smokey/blob/main/features/account_api.feature
+
+#### Changing the Pact tests
+
+If you make changes to the API, you'll need to update the Pact tests.
+
+A Pact test has two parts:
+
+1. The consumer test (defined [in gds-api-adapters][]), which:
+   - specifies the state it expects the provider to be in
+   - gives a request to make
+   - and a response to match against
+
+2. The provider configuration (defined [in this repo][]), which
+   defines all of the provider states referenced by the consumer
+   tests.
+
+See the GOV.UK Developer Docs for [how to update a Pact test][].
+
+[in this repo]: https://github.com/alphagov/account-api/blob/main/spec/service_consumers/pact_helper.rb
+[in gds-api-adapters]: https://github.com/alphagov/gds-api-adapters/blob/master/test/account_api_test.rb
+[how to update a Pact test]: https://docs.publishing.service.gov.uk/manual/pact-broker.html#updating-pact-tests
+
 
 ## Licence
 
--- a/README.md
+++ b/README.md
@@ -3,12 +3,12 @@
 Provides sign in / sign out and attribute storage functionality to
 other GOV.UK applications.
 
-In production this currently uses the [GOV.UK account manager
+In production this app currently uses the [GOV.UK account manager
 prototype][] as an authentication service and the [GOV.UK attribute
 service prototype][] as a user data store.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,85 @@
...
+
+[GOV.UK account manager prototype]: https://github.com/alphagov/govuk-account-manager-prototype
+[GOV.UK attribute service prototype]: https://github.com/alphagov/govuk-attribute-service-prototype
+[Finder Frontend]: https://github.com/alphagov/finder-frontend
+[account-manager]: http://www.login.service.dev.gov.uk/
+[tc-results]: http://finder-frontend.dev.gov.uk/transition-check/results?c[]=living-ie
+[tc-save-results]: http://finder-frontend.dev.gov.uk/transition-check/save-your-results?c%5B%5D=living-ie
+[tc-start]: http://finder-frontend.dev.gov.uk/transition-check/questions
+
+
```
huwd, line 32: Nitpick, double new line?

huwd (approved):
Content looks good and clear.
Nice one!

Would be interested to hear what @jonathanglassman thinks?

jonathanglassman:
> Content looks good and clear.
> Nice one!
> 
> Would be interested to hear what @jonathanglassman thinks?

The API endpoints content looks good. I'm a bit confused about the Requirements for API consumers section - would be good to have a chat with @barrucadu about it.

docs/api.md
```diff
@@ -0,0 +1,248 @@
+# API documentation
+
```
jonathanglassman, line 2: Recommend having an introduction to the API in terms of what the user is trying to achieve. For example:

This API is for GOV.UK services that want to add user sign in / sign out and attribute storage functionality to their applications. When you have integrated this API with your service, you can sign in and authenticate your users' identities on GOV.UK.

docs/api.md
```diff
@@ -0,0 +1,248 @@
...
+  - [`GET /api/oauth2/sign-in`](#get-apioauth2sign-in)
+  - [`POST /api/oauth2/callback`](#post-apioauth2callback)
+  - [`POST /api/oauth2/state`](#post-apioauth2state)
+  - [`GET /api/attributes`](#get-apiattributes)
+  - [`PATCH /api/attributes`](#patch-apiattributes)
+  - [`GET /api/transition-checker-email-subscription`](#get-apitransition-checker-email-subscription)
+  - [`POST /api/transition-checker-email-subscription`](#post-apitransition-checker-email-subscription)
+- [Healthcheck](#healthcheck)
+
+## Nomenclature
```
jonathanglassman, line 19: Can we use a more plain English word, for example, definitions or defined terms?

docs/api.md
```diff
@@ -0,0 +1,248 @@
...
+**Request:** ([see VCL](https://github.com/alphagov/govuk-cdn-config/blob/2ade45759be947a28b1225aee085311430bcbecf/vcl_templates/www.vcl.erb#L339-L340))
+
+- `GOVUK-Account-Session`: set by our CDN to the value of the user's session cookie.  This is the user's Session Identifier.
+
+**Response:** ([see VCL](https://github.com/alphagov/govuk-cdn-config/blob/2ade45759be947a28b1225aee085311430bcbecf/vcl_templates/www.vcl.erb#L442-L455))
+
+- `GOVUK-Account-Session`: sent to our CDN to update the user's session cookie.
+- `GOVUK-Account-End-Session`: sent to our CDN to delete the user's session cookie.
+
+Every app which uses this API is responsible for using the custom
```
jonathanglassman, line 46: Recommend making more active and focused on the service. For example:

"If your service's app uses this API, you are responsible for..."

docs/api.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Pass the Session Identifier to Access Restricted endpoints
+
+Some of these endpoints are Access Restricted, and require the
+`GOVUK-Account-Session` request header to be set.
+
+The [gds-api-adapters][] methods for these endpoints have a required
+`govuk_account_session:` parameter, which must be set to the latest
+Session Identifier.
+
```
jonathanglassman, line 58: It looks to me like I, the developer on the service team who wants to integrate this service with the API, need to do two things here:

- set the `GOVUK-Account-Session` request header
- set the required `govuk_account_session:` parameter to the latest session identifier

Is that correct? Does setting one set the other?

Also, how do I set them? What do I actually need to do, and does this documentation need to tell me how to do this? @huwd suggested putting in an example request here.

docs/api.md
```diff
@@ -0,0 +1,248 @@
...
+```ruby
+response.headers["GOVUK-Account-Session"] = "<govuk_account_session value>"
+```
+
+### End the user's session if an Access Restricted endpoint returns a `401: Unauthenticated`
+
+If an Access Restricted endpoint returns a `401: Unauthenticated`
+response, then the user must no longer be considered logged in, and
+the end-session response header must be set when serving the response
```
jonathanglassman, line 77: passive => active

jonathanglassman (changes requested):
API reference looks good. For other sections, have suggested clarifications and asked questions where necessary.

README.md
barrucadu: I normally put a double new line before a h2, though I see I didn't do that for the `Further documentation` heading.

barrucadu:
Thanks for the comments.  @jonathanglassman sorry for the confusion, I don't think I mentioned this anywhere, but this API is specifically for GOV.UK internal use, like [publishing-api](https://github.com/alphagov/publishing-api/), [router-api](https://github.com/alphagov/router-api/), or [email-alert-api](https://github.com/alphagov/email-alert-api/) - it's not something which other service teams will come into contact with.

docs/api.md
barrucadu: Maybe, but "nomenclature" is what other GOV.UK projects are using, eg:

- https://github.com/alphagov/email-alert-api#nomenclature
- https://github.com/alphagov/content-publisher/#nomenclature
- https://github.com/alphagov/support-api/#nomenclature
- https://github.com/alphagov/finder-frontend/#nomenclature

barrucadu: I'll see if I can rework the rest of the doc to not need custom terminology

barrucadu:
I've removed the "Access Restricted" term (editing other sections where needed), reworked the "Requirements for API consumers" section, added an "Example API usage" section, and added missing descriptions to the two `transition-checker-email-subscription` endpoints.

docs/api.md
```diff
@@ -0,0 +1,339 @@
...
+      value: "",
+      domain: "dev.gov.uk",
+      expires: 1.second.ago,
+    }
+  end
+end
+```
+
+
+## API endpoints
```
jonathanglassman, line 187: @barrucadu apologies I forgot to include some feedback on this section. Best practice for API reference documentation is to include the following:

- Endpoints and methods
- Parameters and arguments
- Example requests and responses for each endpoint with example parameters
- Error response codes

This content has the first two. Is it possible to include the example requests and responses, and error response codes? Happy to discuss at any point.

docs/api.md
barrucadu: Added

> This API is for GOV.UK Publishing microservices (the things which make
> up www.gov.uk) to implement personalisation and user session
> management.  It is not for other government services.

barrucadu:
I've reworked the doc a fair bit so I'll resolve the outdated comments.

docs/api.md
barrucadu: :+1: I've added response codes and example requests & responses.

jonathanglassman (approved):

PR #33, opened 2021-03-31
Use correct callback path

barrucadu:
We're logging people in via /transition-check/login/callback currently, whoops.  Well, the fact that that has worked flawlessly just shows that we did the migration without breaking compatibility anywhere.


huwd (approved):
good spot

PR #41, opened 2021-04-14
Encrypt & sign session identifiers

barrucadu:
We're going to introduce a new level of authentication (LoA) to the
account, and so we want to track which LoA a user logged in with so
that the account-api can reject requests where the user isn't
authenticated enough.

This is so that we can:

- require that the user has done (or had remembered) MFA when they use
  the Transition Checker;

- but NOT require MFA to save pages, where it's too heavyweight.

We want to store the LoA in the session header, so we don't need to
query the auth service on every request.  But we don't want the user
to be able to tamper with it, so we now need to sign the header.

For future-proofing, I've gone for storing a JSON object in the
session, rather than a comma-separated string.  It'll be easier to add
new fields this way.

The encoding is:

    (1) urlsafebase64(
    (2)   encrypt(
            {
              access_token: "access_token",
              refresh_token: "refresh_token",
    (3)     }.to_json,
          )
        )

1. We're storing this in a header, so it needs to be a URL-safe value
2. It's encrypted & signed
3. We turn the hash into a JSON string, because encryption operates on
   strings

To avoid booting out all current users, the legacy unsigned sessions
are accepted for now (and are upgraded into signed sessions).  Those
will be removed once we need to have LoA in the session.

Note: the account-api checking LoA is NOT how LoA will be done for
attributes in the Digital Identity-run attribute store.  If they want
to have attributes which need different LoAs to access, they will
implement that themselves.  It would be a security breach waiting to
happen if they asked every service to independently implement access
checks.  This change is *just* for GOV.UK attributes.

But later cards about handling 403 errors and requesting a specific
LoA when authenticating do move us towards being able to work with a
multi-LoA Digital Identity-run attribute store.

---

[Trello card](https://trello.com/c/uLJZAckP/713-sign-the-user-session-identifier)


huwd:
Remind me, what's the expiry time for these legacy sessions?
I'm wondering when the last one will theoretically lapse?

barrucadu:
There is no expiry time, they're session cookies.  I think we'll just drop the legacy sessions when we do https://trello.com/c/G2uYUDuz/716-check-the-level-of-authentication-when-using-attributes-in-the-account-api

app/lib/session_encryptor.rb
```diff
@@ -0,0 +1,64 @@
+class SessionEncryptor
+  KEY_LEN = ActiveSupport::MessageEncryptor.key_len
```
huwd, line 2: this is interesting, so what's the MessageEncryptor here? It's a built in from active support?

Does that mean it's a standard length or does ActiveSupport read this out of the config file?

Would be interested to know what's going on here?

barrucadu:
Pushed a commit moving the base64 ciphertext decoding into `decrypt_string`, so that `decrypt_string` is the inverse of `encrypt_string`.

app/lib/session_encryptor.rb
```diff
@@ -0,0 +1,63 @@
...
+    )
+  end
+
+  def decrypt_session(ciphertext)
+    return if ciphertext.blank?
+
+    plaintext = decrypt_string(ciphertext)
+    return unless plaintext
+
+    hash = JSON.parse(plaintext)
```
huwd, line 23: Stupid question, the keys in this are only going to be string rights?
JSON only ever stores strings... unlike YAML... so there's no chance we end up with symbols here that fail on L26/L27?

app/lib/session_encryptor.rb
barrucadu: Each cipher has a standard length (https://apidock.com/rails/ActiveSupport/MessageEncryptor/key_len/class), the default cipher is AES-256-CBC (https://github.com/rails/rails/blob/2bbb519482dbc7969321719cb515216cf1515bfc/activesupport/lib/active_support/message_encryptor.rb#L87)

app/lib/session_encryptor.rb
barrucadu: Yeah, it's an object with string keys and string values

app/lib/session_encryptor.rb
barrucadu: The crypto here is more or less taken straight from the example in https://apidock.com/rails/ActiveSupport/MessageEncryptor

The only new bits are:

1. The JSON encoding / decoding
2. Storing the salt in the returned value

app/lib/session_encryptor.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def encryptor(salt)
+    key = ActiveSupport::KeyGenerator.new(session_signing_key).generate_key(salt, KEY_LEN)
+    ActiveSupport::MessageEncryptor.new key
+  end
+
+  def generate_salt
+    SecureRandom.hex KEY_LEN
+  end
```
AlanGabbianelli, line 62: Is it worth calling this method `salt` and deleting line 36?

barrucadu: The local variable is also used on line 37, and we want both to get the same value.  But I'll just inline the method.

AlanGabbianelli: I was thinking about memoization but didn't mention it in my previous message:
```
def salt
  @salt ||= SecureRandom.hex(KEY_LEN)
end
```

app/lib/session_encryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def encrypt_string(plaintext)
+    salt = SecureRandom.hex KEY_LEN
+    ciphertext = encryptor(salt).encrypt_and_sign(plaintext)
+    Base64.urlsafe_encode64("#{salt}$$#{ciphertext}")
+  end
+
+  def decrypt_string(ciphertext)
+    bits = Base64.urlsafe_decode64(ciphertext).split("$$")
+    return nil unless bits.length == 2
```
AlanGabbianelli, line 43: I assume we expect only 2 bits because we know that `"$$"` is never present in `salt` nor in `ciphertext`. I get that for `salt` because it's a random hexadecimal string. Is that the case for `ciphertext` too?

barrucadu: Yes, `ciphertext` is base64 encoded: https://apidock.com/rails/ActiveSupport/MessageEncryptor

AlanGabbianelli: Ah, [I see](https://en.wikipedia.org/wiki/Base64#Base64_table)!

app/lib/session_encryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+protected
+
+  attr_reader :session_signing_key
+
+  def encrypt_string(plaintext)
+    salt = SecureRandom.hex KEY_LEN
+    ciphertext = encryptor(salt).encrypt_and_sign(plaintext)
+    Base64.urlsafe_encode64("#{salt}$$#{ciphertext}")
```
AlanGabbianelli, line 38: Why do we want to encode the salt too? Isn't that only needed to create the ciphertext?

app/lib/session_encryptor.rb
AlanGabbianelli: That's fair

app/lib/session_encryptor.rb
barrucadu: The salt is also needed for decryption, so it needs to be stored in the user's session too.

AlanGabbianelli: Never mind, I see that it's used when decrypting.

huwd (approved):

app/lib/session_encryptor.rb
AlanGabbianelli: We wrote at the same time 😅

AlanGabbianelli (approved):

PR #42, opened 2021-04-15
Store the level of authentication in the session 

barrucadu:
This should be deployed after the change to the account manager, as it
throws an error if the expected level_of_authentication parameter
isn't returned.

The use of LOWEST_LEVEL_OF_AUTHENTICATION in the SessionHeaderHelper
means that current sessions will be upgraded to level0 sessions.  This
won't be enough to use the Transition Checker, so they will end up
needing to reauthenticate, but it is enough to use the bookmark
experiment.

---

[Trello card](https://trello.com/c/yCLEBvxW/714-store-the-level-of-authentication-in-the-users-session)

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -32,11 +32,12 @@ class AuthenticationController < ApplicationController
 
     render json: {
       govuk_account_session: to_account_session(
-        access_token: oauth_response[:access_token],
-        refresh_token: oauth_response[:refresh_token],
+        access_token: oauth_response.fetch(:access_token),
+        refresh_token: oauth_response.fetch(:refresh_token),
+        level_of_authentication: oauth_response.fetch(:result).fetch("level_of_authentication"),
       ),
       redirect_path: redirect_path,
-      ga_client_id: oauth_response[:result]["_ga"],
+      ga_client_id: oauth_response.fetch(:result)["_ga"],
     }
   rescue OidcClient::OAuthFailure
     head :unauthorized
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -31,7 +31,10 @@ class AuthenticationController < ApplicationController
     auth_request.delete
 
     render json: {
-      govuk_account_session: to_account_session(oauth_response[:access_token], oauth_response[:refresh_token]),
+      govuk_account_session: to_account_session(
+        access_token: oauth_response[:access_token],
+        refresh_token: oauth_response[:refresh_token],
+      ),
       redirect_path: redirect_path,
       ga_client_id: oauth_response[:result]["_ga"],
     }
--- a/app/helpers/session_header_helper.rb
+++ b/app/helpers/session_header_helper.rb
@@ -1,10 +1,11 @@
 module SessionHeaderHelper
-  def to_account_session(access_token:, refresh_token:)
+  def to_account_session(access_token:, refresh_token:, level_of_authentication:)
     SessionEncryptor
       .new(session_signing_key: Rails.application.secrets.session_signing_key)
       .encrypt_session(
         access_token: access_token,
         refresh_token: refresh_token,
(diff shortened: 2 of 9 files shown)


app/helpers/session_header_helper.rb
```diff
@@ -1,10 +1,15 @@
...
-  def to_account_session(access_token, refresh_token)
+  LOWEST_LEVEL_OF_AUTHENTICATION = "level0"
+
+  def to_account_session(access_token:, refresh_token:, level_of_authentication:)
     SessionEncryptor
       .new(session_signing_key: Rails.application.secrets.session_signing_key)
       .encrypt_session(
         access_token: access_token,
         refresh_token: refresh_token,
+        level_of_authentication: level_of_authentication || LOWEST_LEVEL_OF_AUTHENTICATION,
```
AlanGabbianelli, line 12: Calling `to_account_session` without `level_of_authentication` would raise an ArgumentError so `LOWEST_LEVEL_OF_AUTHENTICATION` wouldn't be used. It could be moved to the method definition:
```ruby
def to_account_session(access_token:, refresh_token:, level_of_authentication: LOWEST_LEVEL_OF_AUTHENTICATION)
```

```ruby
[10] pry(main)> def bla(a:, b:)
[10] pry(main)*   { foo: a, bar: b || "default" }
[10] pry(main)* end
=> :bla
[11] pry(main)> bla(a: "foo")
ArgumentError: missing keyword: :b
from (pry):16:in `bla'
[12] pry(main)> def bla(a:, b: "default")
[12] pry(main)*   { foo: a, bar: b }
[12] pry(main)* end
=> :bla
[13] pry(main)> bla(a: "foo")
=> {:foo=>"foo", :bar=>"default"}
```

AlanGabbianelli: Or are we expecting to always pass it and just give a default when the value is `nil`? 🤔

barrucadu: > Or are we expecting to always pass it and just give a default when the value is nil?

Yes, exactly.  Existing sessions won't have a `level_of_authentication`, so they'll pass `nil`.

Though, on reflection, I think it'll be better to handle that case in `decrypt_session`.  I'll push a fixup.

app/controllers/authentication_controller.rb
```diff
@@ -32,11 +32,12 @@ def callback
...
       govuk_account_session: to_account_session(
-        access_token: oauth_response[:access_token],
-        refresh_token: oauth_response[:refresh_token],
+        access_token: oauth_response.fetch(:access_token),
+        refresh_token: oauth_response.fetch(:refresh_token),
+        level_of_authentication: oauth_response.fetch(:result).fetch("level_of_authentication"),
       ),
       redirect_path: redirect_path,
-      ga_client_id: oauth_response[:result]["_ga"],
+      ga_client_id: oauth_response.fetch(:result)["_ga"],
```
AlanGabbianelli, line 40: Is this because with want a KeyError if `:result` isn't in the oauth_response hash?

barrucadu: Yeah, a missing `oauth_response[:access_token]`, `:response_token`, or `:result` would be a bug, so I thought to use `.fetch` to make that really clear.

AlanGabbianelli: Is that not as important in `app/controllers/attributes_controller.rb`?

barrucadu: I'm actually wondering about refactoring away the `to_account_session`s in other places, as it's only needed if the OAuth tokens change - which they don't always.  But I'm not sure if that would just add too much complexity.

AlanGabbianelli: Would that be in a subsequent PR?

barrucadu: I think so, I want to ruminate on it a bit.

AlanGabbianelli (approved):

PR #43, opened 2021-04-15
Validate attributes and flag where they're stored

barrucadu:
This PR adds a config file holding attribute details, recording whether an attribute is stored locally or not, and adds validation to the AttributesController to reject unknown attributes.

I've also chosen to return a rejection error as an [RFC 7807 "Problem Detail" object](https://tools.ietf.org/html/rfc7807) because

1. we'll need to return some sort of structured error when we implement levels-of-authentication (to communicate back to the caller which LoA to reauthenticate the user at), and
2. it emphasises having online documentation for each distinct error type, which I think is really nice.

This PR isn't a proposal to standardise GOV.UK on using the RFC; but it would be good to be consistent within this one app at least.

---

[Trello card](https://trello.com/c/0kDToZ6G/709-validate-attribute-names-in-the-account-api)

### DIFF
--- a/app/controllers/attributes_controller.rb
+++ b/app/controllers/attributes_controller.rb
@@ -6,30 +6,23 @@ class AttributesController < ApplicationController
 
     return unless validate_attributes(attribute_names)
 
-    access_token = @govuk_account_session[:access_token]
-    refresh_token = @govuk_account_session[:refresh_token]
-
     values = attribute_names.each_with_object({}) do |name, values_hash|
       if user_attributes.stored_locally? name
         raise NotImplementedError
       else
         oauth_response = OidcClient.new.get_attribute(
           attribute: name,
-          access_token: access_token,
-          refresh_token: refresh_token,
+          access_token: @govuk_account_session[:access_token],
+          refresh_token: @govuk_account_session[:refresh_token],
         )
-        access_token = oauth_response[:access_token]
-        refresh_token = oauth_response[:refresh_token]
+        @govuk_account_session[:access_token] = oauth_response[:access_token]
+        @govuk_account_session[:refresh_token] = oauth_response[:refresh_token]
         values_hash[name] = oauth_response[:result]
       end
     end
 
     render json: {
-      govuk_account_session: to_account_session(
-        access_token: access_token,
-        refresh_token: refresh_token,
-        level_of_authentication: @govuk_account_session[:level_of_authentication],
-      ),
+      govuk_account_session: account_session_header_value,
       values: values.compact,
     }
   rescue OidcClient::OAuthFailure
@@ -41,9 +34,6 @@ class AttributesController < ApplicationController
 
     return unless validate_attributes(attributes.keys)
 
-    access_token = @govuk_account_session[:access_token]
-    refresh_token = @govuk_account_session[:refresh_token]
-
     local_attributes = attributes.select { |name| user_attributes.stored_locally? name }
     remote_attributes = attributes.reject { |name| user_attributes.stored_locally? name }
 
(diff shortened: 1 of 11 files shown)


app/controllers/attributes_controller.rb
```diff
@@ -29,16 +35,48 @@ def show
...
-      attributes: attributes,
-      access_token: @govuk_account_session[:access_token],
-      refresh_token: @govuk_account_session[:refresh_token],
-    )
+    return unless validate_attributes(attributes.keys)
+
+    access_token = @govuk_account_session[:access_token]
+    refresh_token = @govuk_account_session[:refresh_token]
+
+    local_attributes = attributes.select { |name, _| user_attributes.stored_locally? name }
```
huwd, line 43: Why do you need the second `_` param here?

app/controllers/attributes_controller.rb
```diff
@@ -29,16 +35,48 @@ def show
...
+    remote_attributes = attributes.reject { |name, _| user_attributes.stored_locally? name }
```
huwd, line 44: and here?

app/controllers/attributes_controller.rb
barrucadu: Aha

app/lib/user_attributes.rb
```diff
@@ -0,0 +1,42 @@
+class UserAttributes
+  CONFIG_KEYS = %w[is_stored_locally].freeze
+
+  attr_reader :attributes
+
+  def initialize
+    @attributes = YAML.safe_load(File.read(Rails.root.join("config/user_attributes.yml"))).with_indifferent_access
+
+    unless Rails.env.production?
```
huwd, line 9: Are you targeting the test env with with?
This will mean we load these for dev too won't it?

app/lib/user_attributes.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def errors
+    attributes.each_with_object({}) do |(name, config), errors|
+      config ||= {}
+
+      missing_keys = CONFIG_KEYS.reject { |key| config.keys.include? key }
+      unknown_keys = config.keys.reject { |key| CONFIG_KEYS.include? key }
+      invalid_keys = [
+        [nil, true, false].include?(config["is_stored_locally"]) ? nil : :is_stored_locally,
```
huwd, line 30: nitpick, quite a lot going on here, wonder if it's easier to read if this is off in it's own method?

barrucadu: Is afcc7b44069836d3bc8f04cd3e083fc63388ea67 better?

docs/api.md
```diff
@@ -498,6 +502,35 @@ Response:
 ```
 
 
+## API errors
+
+API errors are returned as an [RFC 7080][] "Problem Detail" object, in
```
barrucadu, line 507: Should say "RFC 7807", not "RFC 7080".

spec/requests/attributes_controller_spec.rb
```diff
@@ -3,10 +3,10 @@
 
   let(:headers) { { "Content-Type" => "application/json", "GOVUK-Account-Session" => placeholder_govuk_account_session } }
 
-  let(:attribute_name1) { "name" }
+  let(:attribute_name1) { "foo" }
   let(:attribute_value1) { { "some" => "complex", "value" => 42 } }
 
-  let(:attribute_name2) { "name2" }
+  let(:attribute_name2) { "bar" }
```
AlanGabbianelli, line 9: Can we make it clearer that `attribute_name1` and `attribute_name2` need to correspond to the keys in `spec/fixtures/user_attributes.yml` for the tests to pass?

Usually, `foo` and `bar` are used as placeholders when the value is not important. In this case, that isn't entirely true: the value itself isn't meaningful but needs to match another value.

AlanGabbianelli: Something like `fixtures_attribute1` and `fixtures_attribute2` perhaps?

barrucadu: `foo` and `bar` are also used in the pact tests in gds-api-adapters: https://github.com/alphagov/gds-api-adapters/blob/master/test/account_api_test.rb

They weren't meaningful but, with this config file, they now are.

AlanGabbianelli (approved):

spec/requests/attributes_controller_spec.rb
AlanGabbianelli: If the name is meaningful, why don't we use something else that makes that clear?

barrucadu: I've pushed another commit which will let gds-api-adapters be updated: 4c1c4f716d4c1bf08b569cf7d7d1df3b7f545224

AlanGabbianelli (approved):

PR #49, opened 2021-04-26
Refactor the session handling to all be in a single class

barrucadu:
Previously, the session handling was spread across several classes:

- The `ApplicationController` had methods to decode and encode it, delegating to
- The `SessionHeaderHelper`, which detected if it was a current or legacy session decoding and filling in defaults appropriately, which called
- The `SessionEncryptor`, which handled the actual encryption.
- And all the controllers had to keep the access token and refresh token up to date after an OAuth API call.

This PR centralises all of that logic in a new `AccountSession` class.  The `SessionEncryptor` has been partially moved into the `AccountSession`, and partially moved into a new more generic `StringEncryptor` class.  Overall, these changes make the business logic (as opposed to the boilerplate token management) much more clear in the controllers.

I've also added a couple of new tests.


app/lib/string_encryptor.rb
```diff
@@ -0,0 +1,36 @@
...
+    return nil unless bits.length == 2
+
+    salt = bits[0]
+    message = bits[1]
+
+    encryptor(salt).decrypt_and_verify(message)
+  rescue ArgumentError
+    nil
+  rescue ActiveSupport::MessageEncryptor::InvalidMessage
+    nil
```
AlanGabbianelli, line 25: Do we need to explicitly return `nil` here, or is it for readability?

barrucadu: I thought it was more readable, but I'm happy to change it if you think not

app/lib/account_session.rb
```diff
@@ -51,13 +51,45 @@ def serialise
...
+    oidc_do :bulk_set_attributes, { attributes: remote_attributes }
+  end
+
+  def has_email_subscription?
+    oidc_do :has_email_subscription
+  end
+
+  def set_email_subscription(slug)
+    oidc_do :update_email_subscription, { slug: slug }
+  end
```
AlanGabbianelli, line 76: It would be good to see these tested.

app/lib/account_session.rb
```diff
@@ -51,13 +51,45 @@ def serialise
...
       level_of_authentication: level_of_authentication,
     }
   end
 
+  def get_remote_attributes(remote_attributes)
+    remote_attributes.index_with { |name| oidc_do :get_attribute, { attribute: name } }
+  end
+
+  def set_remote_attributes(remote_attributes)
+    return unless remote_attributes.any?
```
AlanGabbianelli, line 65: Just nitpicking but I find `if` much more easy to mentally parse than `unless` so perhaps we can change this to `return if remote_attributes.none?`

app/lib/account_session.rb
```diff
@@ -51,13 +51,45 @@ def serialise
...
-      access_token: access_token,
-      refresh_token: refresh_token,
+      access_token: @access_token,
+      refresh_token: @refresh_token,
       level_of_authentication: level_of_authentication,
     }
   end
 
+  def get_remote_attributes(remote_attributes)
+    remote_attributes.index_with { |name| oidc_do :get_attribute, { attribute: name } }
```
AlanGabbianelli, line 61: I didn't know `index_with`, nice one!

app/lib/account_session.rb
barrucadu: They're covered by the request specs, do you think they also need tests just for this class?

app/lib/account_session.rb
barrucadu: It's actually a rubocop suggestion, if your `foo.each_with_object({})` is sufficiently simple

app/controllers/transition_checker_email_subscription_controller.rb
```diff
@@ -2,29 +2,18 @@ class TransitionCheckerEmailSubscriptionController < ApplicationController
...
-      refresh_token: @govuk_account_session.refresh_token,
-    )
-    @govuk_account_session.access_token = oauth_response[:access_token]
-    @govuk_account_session.refresh_token = oauth_response[:refresh_token]
+    has_subscription = @govuk_account_session.has_email_subscription?
 
     render json: {
       govuk_account_session: @govuk_account_session.serialise,
-      has_subscription: oauth_response[:result],
+      has_subscription: has_subscription,
```
AlanGabbianelli: We can shave off another line here if we remove the local variable and just use `has_subscription: @govuk_account_session.has_email_subscription?,`. Not so much to have one less line, but because we only use it here.

barrucadu: Actually we can't, because `has_email_subscription?` might refresh the access token, meaning the result of `@govuk_account_session.serialise` will change (unless we start storing tokens in the database rather than the session).  Do you have any suggestions for how that could be made more explicit?

app/controllers/attributes_controller.rb
```diff
@@ -1,28 +1,20 @@
...
-        )
-        @govuk_account_session[:access_token] = oauth_response[:access_token]
-        @govuk_account_session[:refresh_token] = oauth_response[:refresh_token]
-        values_hash[name] = oauth_response[:result]
-      end
-    end
+    local_attributes = attribute_names.select { |name| user_attributes.stored_locally? name }
+    remote_attributes = attribute_names.reject { |name| user_attributes.stored_locally? name }
+
+    raise NotImplementedError if local_attributes.any?
```
AlanGabbianelli, line 12: Both in `show` and `update` we fetch the attributes, validate them, separate them in local and remote, raise an error if we have local ones. I wonder if we can extract some of this logic from the two actions into private methods.

app/controllers/attributes_controller.rb
```diff
@@ -6,20 +6,12 @@ def show
...
-        @govuk_account_session.refresh_token = oauth_response[:refresh_token]
-        values_hash[name] = oauth_response[:result]
-      end
-    end
+    local_attributes = attribute_names.select { |name| user_attributes.stored_locally? name }
+    remote_attributes = attribute_names.reject { |name| user_attributes.stored_locally? name }
+
+    raise NotImplementedError if local_attributes.any?
+
+    values = @govuk_account_session.get_remote_attributes(remote_attributes)
```
AlanGabbianelli, line 14: Same as `has_subscription`. We only use `values` once so we could avoid a local variable.

app/lib/string_encryptor.rb
AlanGabbianelli: I asked because I was wondering if after a `rescue` the default return value wasn't `nil` (as it usually is) and we needed it explicitly. Even if it is, I think there's value in being explicit after rescuing the errors.

app/lib/account_session.rb
AlanGabbianelli: I think having them unit tested would be good because we rely on them in the controllers and they rely on a private method. It would also document them and make it clearer to see what they do and what they return.

app/controllers/transition_checker_email_subscription_controller.rb
AlanGabbianelli: Ah, that's interesting. I suppose that's because of
`@access_token = oauth_response[:access_token]` in `oidc_do`, which depends on OpenID Connect's behaviour. I'm not sure how we can make this more explicit in the code without writing comments 🤔

app/controllers/attributes_controller.rb
AlanGabbianelli: Never mind, you already said why this isn't possible in the previous comment.

app/controllers/attributes_controller.rb
barrucadu: How's 1ac9f362be1876108fdea1b466dd2e3f6f846cc5?

AlanGabbianelli: Looks good, the only thing I don't understand is why `get_attributes_from_params` returns `[local_attributes, remote_attributes]` instead of just `remote_attributes`. We already know there won't be any at that point because we have `raise NotImplementedError if local_attributes.any?` above.

Returning only `remote_attributes`, we could have `remote_attributes = get_attributes_from_params(...)` instead of `_local_attributes, remote_attributes = get_attributes_from_params(...)`. Or am I missing something?

app/controllers/application_controller.rb
```diff
@@ -12,9 +12,9 @@ def fetch_govuk_account_session
     head :unauthorized unless @govuk_account_session
   end
 
-protected
+private
```
AlanGabbianelli, line 17: 👍

app/controllers/attributes_controller.rb
barrucadu: Ok, I'll change it to just return `remote_attributes`.  We will have `local_attributes` at some point, but not yet.

app/lib/account_session.rb
barrucadu: Ok, I've added tests in d25029373d8ef47b2c1e1972d7e0a64549740797

app/controllers/transition_checker_email_subscription_controller.rb
barrucadu: I've changed the code to throw an error if you try to make an OAuth call after serialising the session, which will make improper uses clearly visible at least: 940aaf43f4ec7fdb1fb87eac676d3a6e51e8a538

AlanGabbianelli: That's a good improvement 👍

app/lib/account_session.rb
AlanGabbianelli: Nice 👍

AlanGabbianelli (approved):

PR #55, opened 2021-04-28
Enforce levels of authentication, gated behind a feature flag

barrucadu:
We can't switch this feature on until we have a way to authenticate a user to a specific level (there's a separate card for that) and then update finder-frontend to gracefully handle 403 errors.

Something I'm not entirely satisfied with in this PR is how the level of authentication stored in the session is a string, but really for comparison purposes we want an integer:

- Strings are good because it means that we're going to be using the same level names as Digital Identity, reducing the potential for confusion.
- Integers are good because we need a total ordering, which ideally is pretty efficient to compute, on the levels.

Having both is perhaps not so good, but I think the advantage of keeping around the original level in the session is good, because it means that if Digital Identity update their levels, we can handle that gracefully, as we already have a level-to-integer conversion going on.

---

[Trello card](https://trello.com/c/G2uYUDuz/716-check-the-level-of-authentication-when-using-attributes-in-the-account-api)

### DIFF
--- a/app/controllers/attributes_controller.rb
+++ b/app/controllers/attributes_controller.rb
@@ -2,7 +2,10 @@ class AttributesController < ApplicationController
   before_action :require_govuk_account_session!
 
   def show
-    remote_attributes = get_attributes_from_params(params.fetch(:attributes))
+    remote_attributes = get_attributes_from_params(
+      params.fetch(:attributes),
+      permission_level: :get,
+    )
 
     values = @govuk_account_session.get_remote_attributes(remote_attributes)
 
@@ -15,7 +18,11 @@ class AttributesController < ApplicationController
   end
 
   def update
-    remote_attributes = get_attributes_from_params(params.fetch(:attributes).permit!.to_h, is_hash: true)
+    remote_attributes = get_attributes_from_params(
+      params.fetch(:attributes).permit!.to_h,
+      permission_level: :set,
+      is_hash: true,
+    )
 
     @govuk_account_session.set_remote_attributes(remote_attributes)
 
@@ -28,11 +35,17 @@ class AttributesController < ApplicationController
 
 private
 
-  def get_attributes_from_params(attributes, is_hash: false)
+  def get_attributes_from_params(attributes, permission_level:, is_hash: false)
     attribute_names = is_hash ? attributes.keys : attributes
 
     unknown_attributes = attribute_names.reject { |name| user_attributes.defined? name }
-    raise ApiError::UnknownAttributeNames, unknown_attributes if unknown_attributes.any?
+    raise ApiError::UnknownAttributeNames, { attributes: unknown_attributes } if unknown_attributes.any?
+
+    forbidden_attributes = attribute_names.reject { |name| user_attributes.has_permission_for? name, permission_level, @govuk_account_session }
+    if forbidden_attributes.any?
+      needed_level_of_authentication = forbidden_attributes.map { |name| user_attributes.level_of_authentication_for name, permission_level }.max
+      raise ApiError::LevelOfAuthenticationTooLow, { attributes: forbidden_attributes, needed_level_of_authentication: needed_level_of_authentication }
+    end
 
     local_attributes = attributes.select { |name| user_attributes.stored_locally? name }
     remote_attributes = attributes.reject { |name| user_attributes.stored_locally? name }
--- a/app/controllers/attributes_controller.rb
+++ b/app/controllers/attributes_controller.rb
@@ -41,10 +41,12 @@ private
     unknown_attributes = attribute_names.reject { |name| user_attributes.defined? name }
     raise ApiError::UnknownAttributeNames, { attributes: unknown_attributes } if unknown_attributes.any?
 
-    forbidden_attributes = attribute_names.reject { |name| user_attributes.has_permission_for? name, permission_level, @govuk_account_session }
-    if forbidden_attributes.any?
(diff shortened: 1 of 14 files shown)


app/lib/user_attributes.rb
```diff
@@ -40,6 +41,24 @@ def errors
         invalid_keys << "is_stored_locally"
       end
 
+      permissions = config["permissions"]
+      if permissions
+        missing_keys.concat(PERMISSION_KEYS.reject { |key| permissions.keys.include? key }.map { |key| "permissions.#{key}" })
```
huwd, line 54: humm, there a couple of enumerable operations on these lines, which makes them quite long.

Nitpick so take it or leave it, butperahps these could be split off into a method?

So:
`missing_keys.concat(missing_keys_with_permission_names(permissions)`

app/lib/user_attributes.rb
```diff
@@ -40,6 +41,24 @@ def errors
...
+        unknown_keys.concat(permissions.keys.reject { |key| PERMISSION_KEYS.include? key }.map { |key| "permissions.#{key}" })
+
+        non_integral_values = permissions.keys.reject { |key| permissions[key].is_a? Integer }.map { |key| "permissions.#{key}" }
```
huwd, line 57: Humm.. is there a clearer name than this?
I think it's the most correct, but it took me a moment to realise this was integral as in integer and not as in integrated

app/lib/api_error/level_of_authentication_too_low.rb
```diff
@@ -0,0 +1,27 @@
...
+    def initialize(attributes:, needed_level_of_authentication:)
+      @attributes = attributes
+      @needed_level_of_authentication = LevelOfAuthentication.integer_to_name needed_level_of_authentication
+    end
+
+    def status_code
+      :forbidden
+    end
+
+    def detail
```
huwd, line 12: nice!

app/lib/api_error/level_of_authentication_too_low.rb
```diff
@@ -0,0 +1,27 @@
...
+      I18n.t(
+        "errors.level_of_authentication_too_low.detail",
+        attributes: @attributes.join(", "),
+        needed_level_of_authentication: @needed_level_of_authentication,
+      )
+    end
+
+    def extra_detail
+      {
+        attributes: @attributes,
```
huwd, line 22: When would I want this one, and when would I want the comma separated string from L14?

app/controllers/attributes_controller.rb
```diff
@@ -28,11 +35,19 @@ def update
 
 private
 
-  def get_attributes_from_params(attributes, is_hash: false)
+  def get_attributes_from_params(attributes, permission_level:, is_hash: false)
```
huwd, line 38: Permission levels must be either `:get` `:set` or `:check` right?
Is it work putting any kind fo guard clause up here with an error message for an invalid permission_level?

I guess that would be caught with `ApiError::LevelOfAuthenticationTooLow` if i passed this `:update` or something?

app/lib/api_error/level_of_authentication_too_low.rb
barrucadu: The `detail` string is human readable, the `extra_detail` is machine readable

app/controllers/attributes_controller.rb
barrucadu: It'll throw an error because of the fetch in

```
  def level_of_authentication_for(name, permission_level)
    attributes.fetch(name)[:permissions].fetch(permission_level)
  end
```

app/lib/user_attributes.rb
barrucadu: How about `non_integer_values`?

spec/lib/user_attributes_spec.rb
```diff
@@ -34,7 +34,7 @@
       end
 
       context "when a key has an unexpected value" do
-        let(:foo_properties) { { "is_stored_locally" => "truee", "permissions" => foo_permissions } }
+        let(:foo_properties) { { "is_stored_locally" => "banana", "permissions" => foo_permissions } }
```
huwd, line 37: :D thanks! My brain has to work much harder to confuse me about that one now :D

spec/lib/user_attributes_spec.rb
```diff
@@ -2,4 +2,40 @@
...
+    context "when an unexpected key is present" do
+      let(:bar_properties) { { "is_stored_loally" => true } }
+
+      it "rejects" do
+        expect(errors).to eq({ "bar" => { missing_keys: %w[is_stored_locally], unknown_keys: %w[is_stored_loally] } })
+      end
+    end
+
+    context "when a key has an unexpected value" do
+      let(:foo_properties) { { "is_stored_locally" => "truee" } }
```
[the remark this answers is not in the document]
huwd: [resolved](https://github.com/alphagov/account-api/pull/55/commits/3a67b29657dda1c14dde782142f22d9a0d87ca41)

app/lib/user_attributes.rb
barrucadu: I've condensed it a bit in 04d8265d2e7d94ebaa91fa2d75cdd20d2d729683

app/lib/user_attributes.rb
huwd: yeah that works for me!

app/lib/user_attributes.rb
huwd: looks great, thanks for taking the time to do that

huwd (approved):

PR #57, opened 2021-04-29
Implement support for local attributes

barrucadu:
We don't have any of these defined in the config file yet, so this code is essentially dead, but we will do when we start to work on the location experiment.

I decided to create an `OidcUser` model, which the `LocalAttribute`s are connected to, rather than just store the subject identifier in the `LocalAttribute`s directly, because this approach means that if subject identifiers change (which they almost certainly will do when we migrate from our OIDC provider to Digital Identity's) we only have one record per user to update.

### DIFF
--- a/app/controllers/attributes/names_controller.rb
+++ b/app/controllers/attributes/names_controller.rb
@@ -1,10 +1,13 @@
 class Attributes::NamesController < AttributesController
   def show
-    remote_attributes = get_attributes_from_params(
+    local_attributes, remote_attributes = get_attributes_from_params(
       params.fetch(:attributes),
       permission_level: :check,
     )
 
-    render_api_response values: @govuk_account_session.get_remote_attributes(remote_attributes).compact.keys
+    values = @govuk_account_session.get_local_attributes(local_attributes)
+      .merge(@govuk_account_session.get_remote_attributes(remote_attributes))
+
+    render_api_response values: values.compact.keys
   end
 end
--- a/app/controllers/attributes/names_controller.rb
+++ b/app/controllers/attributes/names_controller.rb
@@ -1,13 +1,8 @@
 class Attributes::NamesController < AttributesController
   def show
-    local_attributes, remote_attributes = get_attributes_from_params(
-      params.fetch(:attributes),
-      permission_level: :check,
-    )
+    attributes = params.fetch(:attributes)
+    validate_attributes!(attributes, :check)
 
-    values = @govuk_account_session.get_local_attributes(local_attributes)
-      .merge(@govuk_account_session.get_remote_attributes(remote_attributes))
-
-    render_api_response values: values.compact.keys
+    render_api_response values: @govuk_account_session.get_attributes(attributes).keys
   end
 end
--- a/app/controllers/attributes/names_controller.rb
+++ b/app/controllers/attributes/names_controller.rb
@@ -1,6 +1,5 @@
 class Attributes::NamesController < AttributesController
   def show
-    attributes = params.fetch(:attributes)
     validate_attributes!(attributes, :check)
 
     render_api_response values: @govuk_account_session.get_attributes(attributes).keys
--- a/app/controllers/attributes_controller.rb
+++ b/app/controllers/attributes_controller.rb
@@ -2,21 +2,25 @@ class AttributesController < ApplicationController
   include AuthenticatedApiConcern
 
   def show
-    remote_attributes = get_attributes_from_params(
+    local_attributes, remote_attributes = get_attributes_from_params(
       params.fetch(:attributes),
       permission_level: :get,
     )
 
-    render_api_response values: @govuk_account_session.get_remote_attributes(remote_attributes).compact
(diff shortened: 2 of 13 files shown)


db/migrate/20210429121634_create_oidc_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOidcUsers < ActiveRecord::Migration[6.1]
+  def change
+    create_table :oidc_users do |t|
+      t.string :sub, null: false
```
huwd, line 4: what does `:sub` refer to, in this case?
Is there a more descriptive name we could use?

app/controllers/attributes_controller.rb
```diff
@@ -41,9 +45,7 @@ def get_attributes_from_params(attributes, permission_level:, is_hash: false)
     local_attributes = attributes.select { |name| user_attributes.stored_locally? name }
     remote_attributes = attributes.reject { |name| user_attributes.stored_locally? name }
 
-    raise NotImplementedError if local_attributes.any?
-
-    remote_attributes
+    [local_attributes, remote_attributes]
```
huwd: how are remote attributes sorted?
Are there any expecrations we should maintain here?

LIkes sort these alphabetically at this point to intersperse the local and remote attributes?

app/controllers/attributes_controller.rb
```diff
@@ -2,21 +2,25 @@ class AttributesController < ApplicationController
...
+    local_attributes, remote_attributes = get_attributes_from_params(
       params.fetch(:attributes).permit!.to_h,
       permission_level: :set,
       is_hash: true,
     )
 
+    @govuk_account_session.set_local_attributes(local_attributes)
     @govuk_account_session.set_remote_attributes(remote_attributes)
 
     render_api_response
```
huwd, line 16: Interesting, how come the render API response happens before we select/reject further down?

config/environments/test.rb
```diff
@@ -15,6 +15,7 @@
     Bullet.enable = true
     Bullet.bullet_logger = true
     Bullet.rails_logger = true
+    Bullet.raise = true
```
huwd, line 18: Nice

app/controllers/attributes/names_controller.rb
```diff
@@ -1,13 +1,8 @@
 class Attributes::NamesController < AttributesController
   def show
-    local_attributes, remote_attributes = get_attributes_from_params(
-      params.fetch(:attributes),
-      permission_level: :check,
-    )
+    attributes = params.fetch(:attributes)
+    validate_attributes!(attributes, :check)
```
huwd, line 4: Excellent, this makes fr much eater controllers too.

Could even go further with something like:

by putting `attributes = params.fetch(:attributes)` into the Concern, memoising it to an instance variable and then using a before action?

app/controllers/attributes_controller.rb
```diff
@@ -2,35 +2,23 @@ class AttributesController < ApplicationController
...
   def update
-    local_attributes, remote_attributes = get_attributes_from_params(
-      params.fetch(:attributes).permit!.to_h,
-      permission_level: :set,
-      is_hash: true,
-    )
-
-    @govuk_account_session.set_local_attributes(local_attributes)
-    @govuk_account_session.set_remote_attributes(remote_attributes)
+    attributes = params.fetch(:attributes).permit!.to_h
```
huwd, line 12: Humm, does `permit!` here undemine the strong attributes?
I wonder if its worth rejecting anything that isn't in a known structure?

db/migrate/20210429121634_create_oidc_users.rb
barrucadu: It's the subject identifier, also used here: https://github.com/alphagov/account-api/blob/a53a71fea170d3497edc51b2c20f8568cff3312e/app/controllers/authentication_controller.rb#L36

app/controllers/attributes_controller.rb
barrucadu: What would be the benefit of that?

app/controllers/attributes_controller.rb
barrucadu: I'm not sure what you mean?  The select / reject is done in a method called further up.

app/controllers/attributes/names_controller.rb
barrucadu: Sure

app/controllers/attributes_controller.rb
huwd: Ok fair enough, That's a well point well made!

app/controllers/attributes_controller.rb
huwd: Sorry I wasn't very clear in any of these feedback messages 🤦🏻 
So this one I mean that we're calling `render_api_response` does that not return at this point?

Then we're doing more filtering below that to finally return an array.

I would have thought a render statement would come at the end of a controller method, so I'm a little unsure what's going on here?

huwd (approved):

PR #58, opened 2021-04-30
Add `/api/attributes/names` endpoint

AlanGabbianelli:
This endpoint allows querying if the user has a value stored for an
attribute, without returning the value.

It is very similar to `/api/attributes`, with the difference
that it only returns the name of the attribute instead of the name and
the value.

Trello card: https://trello.com/c/e7JQlXx9/710-add-an-endpoint-to-account-api-to-check-if-an-attribute-is-defined

### DIFF
--- a/app/controllers/attributes/names_controller.rb
+++ b/app/controllers/attributes/names_controller.rb
@@ -1,0 +1,10 @@
+class Attributes::NamesController < AttributesController
+  def show
+    remote_attributes = get_attributes_from_params(
+      params.fetch(:attributes),
+      permission_level: :check,
+    )
+
+    render_api_response values: @govuk_account_session.get_remote_attributes(remote_attributes).compact.keys
+  end
+end
--- a/spec/requests/attributes/names_controller_spec.rb
+++ b/spec/requests/attributes/names_controller_spec.rb
@@ -1,0 +1,126 @@
+RSpec.describe Attributes::NamesController do
+  before do
+    stub_oidc_discovery
+
+    fixture_file = YAML.safe_load(File.read(Rails.root.join("spec/fixtures/user_attributes.yml"))).with_indifferent_access
+    allow(UserAttributes).to receive(:load_config_file).and_return(fixture_file)
+  end
+
+  let(:session_identifier) { placeholder_govuk_account_session }
+  let(:headers) { { "Content-Type" => "application/json", "GOVUK-Account-Session" => session_identifier } }
+
+  # names must be defined in spec/fixtures/user_attributes.yml
+  let(:attribute_name1) { "test_attribute_1" }
+  let(:attribute_name2) { "test_attribute_2" }
+  let(:unknown_attribute_name1) { "this_does_not_exist1" }
+  let(:unknown_attribute_name2) { "this_does_not_exist2" }
+
+  let(:attribute_value1) { "some_value1" }
+  let(:attribute_value2) { "some_value2" }
+
+  let(:status) { 200 }
+  let(:response_body) { JSON.parse(response.body) }
+
+  describe "GET #show" do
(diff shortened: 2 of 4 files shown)


app/controllers/attributes/names_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Attributes::NamesController < AttributesController
+  def show
+    remote_attributes = get_attributes_from_params(
+      params.fetch(:attributes),
+      permission_level: :get,
```
barrucadu, line 5: The permission level here is `:check`, not `:get`

app/controllers/attributes/names_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Attributes::NamesController < AttributesController
+  def show
```
huwd, line 2: I really like this return to restful routes. So +1 for the `:show`

I'm now tinking about how far we might take the descriptive private methods:

```ruby
  def show
  
    render json: {
      values:  get_attribute_values_from_remote_data_store,
    }
  rescue OidcClient::OAuthFailure
    head :unauthorized
  end

private

  def get_attribute_values_from_remote_data_store
      remote_attributes = get_attributes_from_params(
        params.fetch(:attributes),
        permission_level: :check,
      )

    values = @govuk_account_session.get_remote_attributes(remote_attributes)
    values.compact.keys
  end
  ```

Is it any clearer if we moved some of this into private methods?

Comment (barrucadu) [reply app/controllers/attributes/names_controller.rb:2]: That `show` wouldn't be correct, because calling `get_remote_attributes` can change the result of `serialise`.  To protect against silent failure, it'll instead throw an error:
- https://github.com/alphagov/account-api/blob/main/app/lib/account_session.rb#L60
- https://github.com/alphagov/account-api/blob/main/app/lib/account_session.rb#L96

File: spec/requests/attributes/names_controller_spec.rb [hunk 3]
```diff
@@ -0,0 +1,32 @@
...
+  let(:headers) { { "Content-Type" => "application/json", "GOVUK-Account-Session" => session_identifier } }
+
+  # names must be defined in spec/fixtures/user_attributes.yml
+  let(:attribute_name1) { "test_attribute_1" }
+  let(:attribute_value1) { "some_value" }
+
+  let(:params) { { attributes: [attribute_name1] } }
+  let(:status) { 200 }
+
+  describe "GET #show" do
```
barrucadu, line 19: Could we also have a test here for requesting two attributes, where one of them doesn't have a value?

app/controllers/attributes/names_controller.rb
AlanGabbianelli: We had a similar discussion [here](https://github.com/alphagov/account-api/pull/49#discussion_r620193764). Maybe that's a sign that this isn't very clear and we should improve it. Probably not in scope for this PR, but is it possible to make getting attributes not change the result of `serialise`? 🤔

spec/requests/attributes/names_controller_spec.rb
AlanGabbianelli: Definitely! That's a good idea.

app/controllers/attributes/names_controller.rb
barrucadu: We could store the OAuth access and refresh token in the account-api database, rather than the session, but then we have to decide when to expire them (or the database will grow indefinitely) - currently session expiration is entirely a client-side thing.

barrucadu: We've got a potential work-around in slack

barrucadu (dismissed):

barrucadu (approved):

barrucadu (approved):

PR #59, opened 2021-04-30
Create a concern to handle common API functionality

barrucadu:
All authenticated APIs need to respond to an OAuth failure with a
401 (because it means the user's session is invalid) and to include
the serialised session identifier in the response body.

Doing that in each controller method results in a lot of duplication,
and also confusion, as logic like this is subtly wrong:

    render json: {
      foo: @govuk_account_session.go_something_oauthy,
    }

As the serialisation needs to happen *after* any other calls.


AlanGabbianelli (approved):
I really like this approach. Very nice improvement. ⭐🏅

PR #62, opened 2021-05-05
Add new provider_state for pact tests

AlanGabbianelli:
This adds a new provider_state to capture the case when an attribute
name is present but it has no value.

This is used in https://github.com/alphagov/gds-api-adapters/pull/1060

Trello card: https://trello.com/c/e7JQlXx9/710-add-an-endpoint-to-account-api-to-check-if-an-attribute-is-defined

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -118,6 +118,12 @@ Pact.provider_states_for "GDS API Adapters" do
     end
   end
 
+  provider_state "there is a valid user session, with an attribute called 'foo' that has no value" do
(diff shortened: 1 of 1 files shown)


barrucadu (approved):

AlanGabbianelli:
You're right, it's not. Thanks for clarifying my doubts about this on slack.
Just to also leave some context here, we said this is not needed because it's effectively covered by [this](https://github.com/alphagov/account-api/blob/add-provider-state-for-pact-tests/spec/service_consumers/pact_helper.rb#L99-L107). Also, we don't want `nil` attribute values.

PR #68, opened 2021-05-12
Take session from cookie when on development

huwd:
This section expects that the Bearer token will always arrive via a
header, however on development we put it in a cookie.

This adds a short method that conditionally reads from there if you are
on dev.

### DIFF
--- a/app/controllers/concerns/authenticated_api_concern.rb
+++ b/app/controllers/concerns/authenticated_api_concern.rb
@@ -4,7 +4,7 @@ module AuthenticatedApiConcern
   included do
     before_action do
       @govuk_account_session = AccountSession.deserialise(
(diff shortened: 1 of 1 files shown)


barrucadu (commented):
account-api always gets it in a header, the cookie is used by frontend apps

huwd:
🤔 what am I missing here then? L24 sees to make a difference on development to what is returned?

The alternative to this would be to copy the value out of the cookie on dev and put it in a header manually if wanted to test an endpoint, it think?

huwd:
How about with the last amendment, it'll now only try the cookie if the header is nil AND we're on local dev?
Otherwise always the header?

barrucadu (approved):

PR #69, opened 2021-05-12
Save a page

huwd:
[Trello card](https://trello.com/c/OvsvoWsO/766-implement-an-api-to-save-a-page-unsave-a-page-check-if-a-page-is-saved-and-return-the-list-of-saved-pages)

Also relevant: [gds-api-adapters#1067](https://github.com/alphagov/gds-api-adapters/pull/1067)

## What

Add save page endpoints allowing an authenticated user to get an index of all their saved pages, as well ass add, delete or query a single saved page.

### DIFF
--- a/app/controllers/saved_pages_controller.rb
+++ b/app/controllers/saved_pages_controller.rb
@@ -6,6 +6,17 @@ class SavedPagesController < ApplicationController
     render_api_response(saved_pages: user.saved_pages.map(&:to_hash))
   end
 
+  # PUT /api/saved_pages/:page_path
+  def update
+    saved_page = SavedPage.find_or_create_by(oidc_user_id: user.id, page_path: params[:page_path])
+
+    if saved_page.persisted?
+      render_api_response(saved_page: saved_page.to_hash)
+    else
+      raise ApiError::CannotSavePage, { page_path: params[:page_path], errors: saved_page.errors }
+    end
+  end
+
 private
 
   def user
--- a/app/controllers/saved_pages_controller.rb
+++ b/app/controllers/saved_pages_controller.rb
@@ -1,0 +1,14 @@
+class SavedPagesController < ApplicationController
+  include AuthenticatedApiConcern
+
+  # GET /api/saved_pages
+  def index
+    render_api_response(saved_pages: user.saved_pages.map(&:to_hash))
+  end
+
+private
+
+  def user
+    @user ||= @govuk_account_session.user
+  end
+end
--- a/app/controllers/saved_pages_controller.rb
+++ b/app/controllers/saved_pages_controller.rb
@@ -1,6 +1,8 @@
 class SavedPagesController < ApplicationController
   include AuthenticatedApiConcern
 
+  before_action :check_saved_page_exists, only: %i[destroy]
+
   # GET /api/saved_pages
   def index
     render_api_response(saved_pages: user.saved_pages.map(&:to_hash))
@@ -17,9 +19,23 @@ class SavedPagesController < ApplicationController
     end
   end
 
+  # DELETE /api/saved_pages/:page_path
+  def destroy
+    saved_page.destroy!
+    head :no_content
+  end
+
 private
 
   def user
     @user ||= @govuk_account_session.user
+  end
+
+  def saved_page
+    @saved_page ||= SavedPage.find_by(oidc_user_id: user.id, page_path: params[:page_path])
+  end
+
+  def check_saved_page_exists
+    head :not_found if saved_page.nil?
   end
 end
--- a/app/controllers/saved_pages_controller.rb
+++ b/app/controllers/saved_pages_controller.rb
@@ -1,11 +1,16 @@
 class SavedPagesController < ApplicationController
   include AuthenticatedApiConcern
 
-  before_action :check_saved_page_exists, only: %i[destroy]
+  before_action :check_saved_page_exists, only: %i[destroy show]
 
   # GET /api/saved_pages
   def index
     render_api_response(saved_pages: user.saved_pages.map(&:to_hash))
+  end
+
+  # GET /api/saved_pages/:page_path
+  def show
+    render_api_response(saved_page: saved_page.to_hash)
   end
 
   # PUT /api/saved_pages/:page_path
--- a/app/models/saved_page.rb
+++ b/app/models/saved_page.rb
@@ -1,9 +1,25 @@
 class SavedPage < ApplicationRecord
   belongs_to :oidc_user
 
+  validates :page_path, uniqueness: { scope: :oidc_user_id }, presence: true
+
+  validate :page_path_is_valid_path
+
   def to_hash
     {
       "page_path" => page_path,
     }
+  end
+
+private
+
+  def page_path_is_valid_path
+    errors.add(:page_path, :invalid_path, message: "must only include URL path") unless is_valid_path?
+  end
+
+  def is_valid_path?
+    page_path&.starts_with?("/") && URI(page_path).path == page_path
+  rescue StandardError
+    false
   end
 end
--- a/app/models/saved_page.rb
+++ b/app/models/saved_page.rb
@@ -1,0 +1,9 @@
+class SavedPage < ApplicationRecord
+  belongs_to :oidc_user
+
+  def to_hash
+    {
+      "page_path" => page_path,
+    }
+  end
+end
--- a/config/locales/errors.en.yml
+++ b/config/locales/errors.en.yml
@@ -9,3 +9,7 @@ en:
       type: https://github.com/alphagov/account-api/blob/main/docs/api.md#unknown-attribute-names
       title: Unknown attribute names
       detail: Attribute names %{attributes} are unknown, have they been added to config/user_attributes.yml?
+    cannot_save_page:
+      type: https://github.com/alphagov/account-api/blob/main/docs/api.md#cannot-save-page
+      title: Cannot save page
+      detail: Cannot save page with path %{page_path}, check it is not blank, and is a well formatted url path.
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -18,6 +18,8 @@ Rails.application.routes.draw do
       get "/names", to: "names#show"
     end
 
+    resources :saved_pages, only: %i[index show update destroy], param: :page_path
+
     get "/transition-checker-email-subscription", to: "transition_checker_email_subscription#show"
     post "/transition-checker-email-subscription", to: "transition_checker_email_subscription#update"
   end
--- a/db/migrate/20210512111851_create_saved_pages.rb
+++ b/db/migrate/20210512111851_create_saved_pages.rb
@@ -1,0 +1,12 @@
+class CreateSavedPages < ActiveRecord::Migration[6.1]
+  def change
+    create_table :saved_pages do |t|
+      t.references :oidc_user, null: false
+      t.string     :page_path, null: false
+
+      t.timestamps
+    end
+
+    add_index :saved_pages, %i[oidc_user_id page_path], unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_04_29_123551) do
+ActiveRecord::Schema.define(version: 2021_05_12_111851) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -38,6 +38,15 @@ ActiveRecord::Schema.define(version: 2021_04_29_123551) do
     t.datetime "created_at", precision: 6, default: -> { "now()" }, null: false
     t.datetime "updated_at", precision: 6, default: -> { "now()" }, null: false
     t.index ["sub"], name: "index_oidc_users_on_sub", unique: true
+  end
+
+  create_table "saved_pages", force: :cascade do |t|
+    t.bigint "oidc_user_id", null: false
+    t.string "page_path", null: false
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["oidc_user_id", "page_path"], name: "index_saved_pages_on_oidc_user_id_and_page_path", unique: true
+    t.index ["oidc_user_id"], name: "index_saved_pages_on_oidc_user_id"
   end
 
   create_table "users", force: :cascade do |t|
--- a/docs/api.md
+++ b/docs/api.md
@@ -5,7 +5,11 @@ up www.gov.uk) to implement personalisation and user session
 management. This API is not for other government services.
 
 - [Nomenclature](#nomenclature)
+  - [Identity provider](#identity-provider)
+  - [Session identifier](#session-identifier)
 - [Requirements for API consumers](#requirements-for-api-consumers)
+  - [Request custom headers](#request-custom-headers)
+  - [Response custom headers](#response-custom-headers)
   - [Update the user's session if a new session identifier is returned](#update-the-users-session-if-a-new-session-identifier-is-returned)
   - [End the user's session if an endpoint returns a `401: Unauthenticated`](#end-the-users-session-if-an-endpoint-returns-a-401-unauthenticated)
   - [Ensure responses are not cached](#ensure-responses-are-not-cached)
--- a/docs/api.md
+++ b/docs/api.md
@@ -24,9 +24,11 @@ management. This API is not for other government services.
   - [`GET /api/transition-checker-email-subscription`](#get-apitransition-checker-email-subscription)
   - [`POST /api/transition-checker-email-subscription`](#post-apitransition-checker-email-subscription)
   - [`GET /api/saved_pages`](#get-apisaved_pages)
+  - [`PUT /api/saved_pages/:page_path`](#put-apisaved_pagespage_path)
 - [API errors](#api-errors)
   - [Level of authentication too low](#level-of-authentication-too-low)
   - [Unknown attribute names](#unknown-attribute-names)
+  - [Page cannot be saved](#page-cannot-be-saved)
 
 ## Nomenclature
 
@@ -611,6 +613,53 @@ Response when a user has saved two pages:
 }
 ```
 
+### `PUT /api/saved_pages/:page_path`
+
+Upsert a saved page in a user's account
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
+
+#### JSON response fields
+
+- `govuk_account_session` *(optional)*
+  - a new session identifier
+- `saved_page`
+  - an object containing the page path of the successfully saved page
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 422 if the page could not be saved (see [error: page cannot be saved](#page-cannot-be-saved))
+- 200 otherwise
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.save_page(
+    page_path: "/guidance/foo",
+    govuk_account_session: "session-identifier",
+)
+```
+
+```json
+{
+    "govuk_account_session": "YWNjZXNzLXRva2Vu.cmVmcmVzaC10b2tlbg==",
+    "saved_page": {
+      "page_path": "/guidance/foo"
+    },
+}
+```
+
 ## API errors
 
 API errors are returned as an [RFC 7807][] "Problem Detail" object, in
@@ -650,3 +699,11 @@ The `attributes` response field lists these.
 - check that you don't have a typo in the attribute names
 - check that the attributes are defined in `config/user_attributes.yml`
 - check that you are running the latest version of account-api
+
+### Page cannot be saved
+
+The page you have specified could not be saved. The errors response field lists the problems.
+
+#### Debugging steps
+
+- check the `errors` returned as an extra detail in the response for specific error messages
--- a/docs/api.md
+++ b/docs/api.md
@@ -23,6 +23,7 @@ management. This API is not for other government services.
   - [`GET /api/attributes/names`](#get-apiattributesnames)
   - [`GET /api/transition-checker-email-subscription`](#get-apitransition-checker-email-subscription)
   - [`POST /api/transition-checker-email-subscription`](#post-apitransition-checker-email-subscription)
+  - [`GET /api/saved_pages`](#get-apisaved_pages)
 - [API errors](#api-errors)
   - [Level of authentication too low](#level-of-authentication-too-low)
   - [Unknown attribute names](#unknown-attribute-names)
@@ -558,6 +559,57 @@ Response:
 }
 ```
 
+### `GET /api/saved_pages`
+
+Returns all a user's saved pages
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### JSON response fields
+
+- `govuk_account_session` *(optional)*
+  - a new session identifier
+- `saved_pages`
+  - an array of pages the user has saved, identified by their page path
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 200 otherwise
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.get_saved_pages(
+    govuk_account_session: "session-identifier",
+)
+```
+
+Response when no pages are saved:
+
+```json
+{
+    "govuk_account_session": "YWNjZXNzLXRva2Vu.cmVmcmVzaC10b2tlbg==",
+    "saved_pages": []
+}
+```
+
+Response when a user has saved two pages:
+
+```json
+{
+    "govuk_account_session": "YWNjZXNzLXRva2Vu.cmVmcmVzaC10b2tlbg==",
+    "saved_pages": [
+      { "page_path": "/government/organisations/government-digital-service" },
+      { "page_path": "/government/organisations/cabinet-office" },
+  ]
+}
+```
 
 ## API errors
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -26,6 +26,7 @@ management. This API is not for other government services.
   - [`GET /api/saved_pages`](#get-apisaved_pages)
   - [`PUT /api/saved_pages/:page_path`](#put-apisaved_pagespage_path)
   - [`DELETE /api/saved_pages/:page_path`](#delete-apisaved_pagespage_path)
+  - [`GET /api/saved_pages/:page_path`](#get-apisaved_pagespage_path)
 - [API errors](#api-errors)
   - [Level of authentication too low](#level-of-authentication-too-low)
   - [Unknown attribute names](#unknown-attribute-names)
@@ -693,6 +694,53 @@ GdsApi.saved_page_api.delete_saved_page(
 ```
 
 Response is status code only.
+
+### `GET /api/saved_pages/:page_path`
+
+Query if a specific path has been saved by the user
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
+
+#### JSON response fields
+
+- `govuk_account_session` *(optional)*
+  - a new session identifier
+- `saved_page`
+  - an object containing the page path of the successfully queried page
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 404 cannot find a page with the provided path
+- 200 otherwise
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.get_saved_page(
+    page_path: "/guidance/bar",
+    govuk_account_session: "session-identifier",
+)
+```
+
+```json
+{
+    "govuk_account_session": "YWNjZXNzLXRva2Vu.cmVmcmVzaC10b2tlbg==",
+    "saved_page": {
+      "page_path": "/guidance/bar"
+    },
+}
+```
 
 ## API errors
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -25,6 +25,7 @@ management. This API is not for other government services.
   - [`POST /api/transition-checker-email-subscription`](#post-apitransition-checker-email-subscription)
   - [`GET /api/saved_pages`](#get-apisaved_pages)
   - [`PUT /api/saved_pages/:page_path`](#put-apisaved_pagespage_path)
+  - [`DELETE /api/saved_pages/:page_path`](#delete-apisaved_pagespage_path)
 - [API errors](#api-errors)
   - [Level of authentication too low](#level-of-authentication-too-low)
   - [Unknown attribute names](#unknown-attribute-names)
@@ -660,6 +661,39 @@ GdsApi.saved_page_api.save_page(
 }
 ```
 
+### `DELETE /api/saved_pages/:page_path`
+
+Remove a saved page from a user's account
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 404 cannot find a page with the provided path
+- 204 successfully deleted
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.delete_saved_page(
+    page_path: "/guidance/bar",
+    govuk_account_session: "session-identifier",
+)
+```
+
+Response is status code only.
+
 ## API errors
 
 API errors are returned as an [RFC 7807][] "Problem Detail" object, in
@@ -693,12 +727,6 @@ level to access.
 
 One or more of the attribute names you have specified are not known.
 The `attributes` response field lists these.
-
-#### Debugging steps
-
-- check that you don't have a typo in the attribute names
-- check that the attributes are defined in `config/user_attributes.yml`
-- check that you are running the latest version of account-api
 
 ### Page cannot be saved
 
--- a/spec/factories/oidc_user_factory.rb
+++ b/spec/factories/oidc_user_factory.rb
@@ -1,0 +1,5 @@
+FactoryBot.define do
+  factory :oidc_user do
+    sequence(:sub) { |n| "user-id-#{n}" }
+  end
+end
--- a/spec/factories/saved_page_factory.rb
+++ b/spec/factories/saved_page_factory.rb
@@ -1,0 +1,6 @@
+FactoryBot.define do
+  factory :saved_page do
+    oidc_user
+    sequence(:page_path) { |n| "/page-path/#{n}" }
+  end
+end
--- a/spec/models/saved_page_spec.rb
+++ b/spec/models/saved_page_spec.rb
@@ -1,0 +1,9 @@
+RSpec.describe SavedPage do
+  describe "to_hash" do
+    let(:saved_page) { FactoryBot.build(:saved_page) }
+
+    it "returns a hash with stringified keys containing the page path" do
+      expect(saved_page.to_hash).to eq({ "page_path" => saved_page.page_path })
+    end
+  end
+end
--- a/spec/models/saved_page_spec.rb
+++ b/spec/models/saved_page_spec.rb
@@ -1,4 +1,53 @@
 RSpec.describe SavedPage do
+  describe "validations" do
+    let(:saved_page) { FactoryBot.build(:saved_page) }
+
+    it "rejects duplicate saved pages, scoped to a user" do
+      saved_page.save!
+      duplicate = saved_page.dup
+      expect(duplicate).not_to be_valid
+      expect(duplicate.errors[:page_path]).to include("has already been taken")
+    end
+
+    it "allows duplicate saved pages, between different users" do
+      saved_page.save!
+      not_a_duplicate = FactoryBot.build(:saved_page, page_path: saved_page.page_path)
+      expect(not_a_duplicate).to be_valid
+    end
+
+    it "rejects blank values as page paths" do
+      [nil, ""].each do |blank_value|
+        saved_page.page_path = blank_value
+        expect(saved_page).not_to be_valid
+        expect(saved_page.errors[:page_path]).to include("can't be blank")
+      end
+    end
+
+    it "rejects paths that do not start with /" do
+      saved_page.page_path = "foo"
+      expect(saved_page).not_to be_valid
+      expect(saved_page.errors[:page_path]).to include("must only include URL path")
+    end
+
+    it "rejects paths that contain spaces" do
+      saved_page.page_path = "/foo bar"
+      expect(saved_page).not_to be_valid
+      expect(saved_page.errors[:page_path]).to include("must only include URL path")
+    end
+
+    it "rejects page paths with parameters" do
+      saved_page.page_path = "/foo/bar?i_am_tracking_identifier=abc123"
+      expect(saved_page).not_to be_valid
+      expect(saved_page.errors[:page_path]).to include("must only include URL path")
+    end
+
+    it "rejects pages with fragment identifiers" do
+      saved_page.page_path = "/guidance/about-the-thing#heading1"
+      expect(saved_page).not_to be_valid
+      expect(saved_page.errors[:page_path]).to include("must only include URL path")
+    end
+  end
+
   describe "to_hash" do
     let(:saved_page) { FactoryBot.build(:saved_page) }
 
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -1,6 +1,10 @@
 require "webmock"
 require "pact/provider/rspec"
 
+def oidc_user
+  OidcUser.find_or_create_by(sub: "user-id")
+end
+
 Pact.configure do |config|
   config.reports_dir = "spec/reports/pacts"
   config.include WebMock::API
@@ -47,7 +51,7 @@ Pact.provider_states_for "GDS API Adapters" do
       id_token: instance_double(
         "OpenIDConnect::ResponseObject::IdToken",
         iss: "http://openid-provider",
-        sub: "user-id",
+        sub: oidc_user.sub,
         aud: "oauth-client",
         exp: 300,
         iat: 0,
@@ -61,7 +65,7 @@ Pact.provider_states_for "GDS API Adapters" do
 
     account_session = AccountSession.new(
       session_signing_key: Rails.application.secrets.session_signing_key,
-      user_id: "user-id",
+      user_id: oidc_user.sub,
       access_token: "access-token",
       refresh_token: "refresh-token",
       level_of_authentication: "level1",
@@ -103,6 +107,30 @@ Pact.provider_states_for "GDS API Adapters" do
       stub_request(:get, "http://openid-provider/v1/attributes/foo").to_return(status: 404)
       stub_request(:get, "http://openid-provider/v1/attributes/test_attribute_1").to_return(status: 404)
       stub_request(:post, "http://openid-provider/v1/attributes").to_return(status: 200)
+    end
+
+    tear_down do
+      SavedPage.destroy_all
+    end
+  end
+
+  provider_state "there is a valid user session, with saved pages" do
+    set_up do
+      FactoryBot.create_list(:saved_page, 2, oidc_user_id: oidc_user.id)
+    end
+
+    tear_down do
+      SavedPage.destroy_all
+    end
+  end
+
+  provider_state "there is a valid user session, with /guidance/some-govuk-guidance saved" do
+    set_up do
+      FactoryBot.create(:saved_page, page_path: "/guidance/some-govuk-guidance", oidc_user_id: oidc_user.id)
+    end
+
+    tear_down do
+      SavedPage.destroy_all
     end
   end
 
--- a/app/lib/api_error/cannot_save_page.rb
+++ b/app/lib/api_error/cannot_save_page.rb
@@ -1,0 +1,25 @@
+module ApiError
+  class CannotSavePage < ApiError::Base
+    attr_reader :page_path, :errors
+
+    def initialize(page_path:, errors:)
+      @page_path = page_path
+      @errors = errors
+    end
+
+    def status_code
+      :unprocessable_entity
+    end
+
+    def detail
+      I18n.t("errors.cannot_save_page.detail", page_path: page_path)
+    end
+
+    def extra_detail
+      {
+        page_path: page_path,
+        errors: errors,
+      }
+    end
+  end
+end
--- a/app/models/oidc_user.rb
+++ b/app/models/oidc_user.rb
@@ -1,3 +1,4 @@
 class OidcUser < ApplicationRecord
   has_many :local_attributes, dependent: :destroy
+  has_many :saved_pages, dependent: :destroy
 end
--- a/spec/requests/saved_pages_spec.rb
+++ b/spec/requests/saved_pages_spec.rb
@@ -1,0 +1,34 @@
+RSpec.describe "Saved pages" do
+  context "when receiving an unauthenticated request" do
+    it "returns unauthorised for GET /api/saved_pages" do
+      get saved_pages_path
+
+      expect(response).to have_http_status(:unauthorized)
+    end
+  end
+
+  context "when receiving an authenticated request" do
+    let(:headers) { { "Content-Type" => "application/json", "GOVUK-Account-Session" => session_identifier } }
+    let(:session_identifier) { placeholder_govuk_account_session(user_id: user.sub) }
+    let(:user) { FactoryBot.create(:oidc_user) }
+
+    describe "GET /api/saved_pages" do
+      it "returns an empty array if there are no saved pages" do
+        get saved_pages_path, headers: headers
+
+        expect(response).to have_http_status(:ok)
+        expect(JSON.parse(response.body)["saved_pages"]).to eq([])
+      end
+
+      it "returns an array of saved_pages if they exist" do
+        list = FactoryBot.create_list(:saved_page, 2, oidc_user_id: user.id)
+        expected_response = list.map(&:to_hash)
+
+        get saved_pages_path, headers: headers
+
+        expect(response).to have_http_status(:ok)
+        expect(JSON.parse(response.body)["saved_pages"]).to eq(expected_response)
+      end
+    end
+  end
+end
--- a/spec/requests/saved_pages_spec.rb
+++ b/spec/requests/saved_pages_spec.rb
@@ -5,6 +5,12 @@ RSpec.describe "Saved pages" do
 
       expect(response).to have_http_status(:unauthorized)
     end
+
+    it "returns unauthorised for PUT /api/saved_pages/:page_path" do
+      put saved_page_path({ page_path: "/my-saved-page-path" })
+
+      expect(response).to have_http_status(:unauthorized)
+    end
   end
 
   context "when receiving an authenticated request" do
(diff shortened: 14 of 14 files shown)


huwd:
Will keep the commits fairly separate for the moment,
Hopefully makes fr an easier review.

I'm consiering etiher squashing them down at the end into a single
`Add save a page feature (#69)`

But that could be a mammoth commit message...

huwd:
Realised I had not added the problem detail objects. went on and did that.

Will tweak the GDS API adaptors

huwd:
damn forgot a couple of API docs, 5 mins will force push again

db/migrate/20210512111851_create_saved_pages.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSavedPages < ActiveRecord::Migration[6.1]
+  def change
+    create_table :saved_pages do |t|
+      t.references :oidc_user, null: false
```
huwd, line 4: This is the one, right? There's also a `User` model, but that seemed to be about something else?

app/controllers/saved_pages_controller.rb
```diff
@@ -8,6 +8,19 @@ def index
     render_api_response(saved_pages: present_pages(@saved_pages))
   end
 
+  # POST /api/saved_pages
+  def create
+    saved_page = SavedPage.create(oidc_user_id: @user.id, page_path: params[:page_path])
+
+    return render_api_response(saved_page: present_page(saved_page)) if saved_page.persisted?
+  rescue ActiveRecord::NotNullViolation
+    head :bad_request
```
huwd: I go back and change these to Error Objects in a later commit!

db/schema.rb
```diff
@@ -40,6 +40,15 @@
     t.index ["sub"], name: "index_oidc_users_on_sub", unique: true
   end
 
+  create_table "saved_pages", force: :cascade do |t|
+    t.bigint "oidc_user_id", null: false
+    t.string "page_path", null: false
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["oidc_user_id", "page_path"], name: "index_saved_pages_on_oidc_user_id_and_page_path", unique: true
+    t.index ["oidc_user_id"], name: "index_saved_pages_on_oidc_user_id"
```
AlanGabbianelli, line 49: This looks good, do you think
`t.index ["page_path"], name: "index_saved_pages_on_page_path"`
would also be needed?

spec/factories/oidc_user_factory.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :oidc_user do
+    sub { "user-id" }
```
AlanGabbianelli, line 3: `sub` is unique so this should be `sequence(:sub) { |n| "user-id-#{n}" }`

spec/factories/saved_page_factory.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :saved_page do
+    oidc_user_id { "user-id" }
```
AlanGabbianelli, line 3: Instead of `oidc_user_id { "user-id" }` you can simply have `oidc_user` here so that when we do `FactoryBot.create(:saved_page)` FactoryBot will also create an associated `oidc_user`.

config/routes.rb
```diff
@@ -18,6 +18,8 @@
       get "/names", to: "names#show"
     end
 
+    resources :saved_pages, only: %i[index show create destroy], param: :page_path
```
AlanGabbianelli, line 21: Nice one, I didn't know this one: https://guides.rubyonrails.org/routing.html#overriding-named-route-parameters

huwd: Yeah I thought that was a cool rails feature too.
The nice bit is, if you run `bundle exec rails routes` it gives you a good print out, and makes clear that we're looking for a `:page_path` instead of a default `:id`!

AlanGabbianelli: Can we maybe clarify this bit in the commit message?
```
Also we're not using an ID here, but instead uniquely identifying a
resource by a page path. So lets name that here.
```

db/migrate/20210512111851_create_saved_pages.rb
huwd: We discussed this in slack, the `User` table is for sso!

db/schema.rb
huwd: I don't _think_ so... let me step through this and check on the why.

For the index path we will look up saved pages by `oicd_user_id` and then return all. So in that case there's no page path, but we may get a nice speed bump from the index.

However there is never a case where we will be looking up a page path independent of a user. It is always an authenticated user looking up either all their saved pages or making a request where we use both their ID and a page path?

I guess there could be a scenario where we start reporting on pages saved and want to look them up, but in that case it could be a rake task, where perhaps speed is not the issue? Or we could add an index then in response to a need?

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def present_page(saved_page)
+    return {} if saved_page.nil?
+
+    { "page_path" => saved_page.page_path }
+  end
+  def set_user
+    @user = OidcUser.find_by(sub: @govuk_account_session.user_id)
+  end
```
AlanGabbianelli, line 59: If we change this to
```
def user
  @user ||= OidcUser.find_by(sub: @govuk_account_session.user_id)
end
```
we can remove the `before_action` and we can use `user` instead of `@user`.

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  end
+
+  context "when receiving an authenticated request" do
+    before do
+      stub_oidc_discovery
+
+      # rubocop:disable RSpec/AnyInstance
+      allow_any_instance_of(OidcClient).to receive(:tokens!).and_return({ access_token: "access-token", refresh_token: "refresh-token" })
+      # rubocop:enable RSpec/AnyInstance
```
AlanGabbianelli, line 34: An alternative to disabling `RSpec/AnyInstance` could be to use an instance double instead of stubbing any instance of `OidcClient`:
```ruby
# outside the before block
let(:oidc_client) { instance_double(OidcClient) }

# ...

# inside the before block
allow(OidcClient).to receive(:new).and_return(oidc_client)
allow(oidc_client).to receive(:tokens!).and_return({ access_token: "access-token", refresh_token: "refresh-token" })
```

This seems to be the approach [suggested by rubocop](https://www.rubydoc.info/gems/rubocop-rspec/1.6.0/RuboCop/Cop/RSpec/AnyInstance).

app/controllers/saved_pages_controller.rb
huwd: that works well, I'll fix that up!

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      FactoryBot.create(:oidc_user)
+    end
+
+    let(:headers) { { "Content-Type" => "application/json", "GOVUK-Account-Session" => session_identifier } }
+    let(:session_identifier) { placeholder_govuk_account_session }
+    let(:user) { OidcUser.find_by(sub: "user-id") }
```
AlanGabbianelli, line 41: Not sure I get the value of creating the user in the before block and then having a let that finds it by `sub`. 🤔
Can't we remove `FactoryBot.create(:oidc_user)` and `let(:user) { OidcUser.find_by(sub: "user-id") }`, and just have `let(:user) { FactoryBot.create(:oidc_user) }`?

spec/requests/saved_pages_request_spec.rb
huwd: great find! I'll amend these and also go back and amend the ones in account_api in a final commit

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    describe "GET /api/saved_pages" do
+      it "returns an empty array if there are no saved pages" do
+        get saved_pages_path, headers: headers
+
+        expect(response).to have_http_status(:ok)
+        expect(JSON.parse(response.body)["saved_pages"]).to eq([])
```
AlanGabbianelli, line 48: Not specifically related to this PR so feel free to ignore this: we use `JSON.parse(response.body)` a lot. It would be nice to have a `response_body` method or a `let(:response_body) { JSON.parse(response.body) }` somewhere that we could use in all our tests.

spec/factories/oidc_user_factory.rb
huwd: Great improvement, thanks!

spec/factories/saved_page_factory.rb
huwd: done!

spec/requests/saved_pages_request_spec.rb
huwd: done!

db/schema.rb
AlanGabbianelli: Sounds good, we'll cross that bridge when we get to it.

config/routes.rb
huwd: amended!

spec/requests/saved_pages_request_spec.rb
huwd: I think you are right, perhaps we can have a tidy up PR after this?

AlanGabbianelli: Sounds good

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,35 @@
+class SavedPagesController < ApplicationController
+  include AuthenticatedApiConcern
+
+  before_action :set_saved_page, only: %i[show destroy]
```
AlanGabbianelli, line 4: I think you left this here while rebasing. This isn't yet used.

app/lib/api_error/saved_page_not_destroyed.rb
```diff
@@ -0,0 +1,21 @@
+module ApiError
+  class SavedPageNotDestroyed < ApiError::Base
+    def initialize(page_path:)
+      @page_path = page_path
+    end
+
+    def status_code
+      :unprocessable_entity
```
barrucadu, line 8: I think this should be a 404: the saved page couldn't be destroyed because it's not found.

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    let(:user) { FactoryBot.create(:oidc_user) }
+
+    before do
+      stub_oidc_discovery
+
+      allow(OidcClient).to receive(:new).and_return(oidc_client)
+      allow(oidc_client).to receive(:tokens!).and_return({ access_token: "access-token", refresh_token: "refresh-token" })
+    end
+
+    describe "GET /api/saved_pages" do
```
AlanGabbianelli, line 43: Should this be in line two and include both contexts ("when receiving an unauthenticated request" and "when receiving an authenticated request")?

app/models/saved_page.rb
```diff
@@ -1,3 +1,33 @@
...
+    errors.add(:page_path, :url_unsafe, message: "must be a URL safe string") if is_url_unsafe_string?
+  end
+
+  def page_path_is_valid_path
+    errors.add(:page_path, :invalid_path, message: "must only include URL path and fragment") if is_invalid_path?
+  end
+
+  def is_invalid_path?
+    return true if page_path.nil?
+    return true if page_path == ""
```
AlanGabbianelli, line 21: This will catch both cases:
```ruby
return true if page_path.blank?
```

app/lib/api_error/saved_page_not_destroyed.rb
huwd: There's [a separate 404 check](https://github.com/alphagov/account-api/pull/69/commits/6a35b1b0e57129feb86fabb8ce5d0ef02248be4f) that is applied as a before action to any route where a page needs to be identified (`show` and `destroy` currently)

This check is more a piece of defensive programming, I can't figure out how anyone would ever end up on this branch, but broadly:
- if they supply a good path
- and they are authenticated
- and yet when we call destroy on the model... it did not work

Then we show the saved page not destroyed error.

Both that and SavedPageNotPersisted shouldn't ever crop up, but if they do it'd be worth us checking those out in sentry!

barrucadu: Aha, I see.  Won't `destroy!` raise an exception if something goes wrong though?  So this bit of defensive code will never be reached?

- If the page doesn't exist, we return a 404
- If it does, and we delete it, we return a 204
- If it does, and deleting it fails, an exception is thrown and we return a 500

spec/requests/saved_pages_request_spec.rb
huwd: So as in why did i group these by auth/unauth _then_ the end points and not the other way around?

Mostly I figure the two fundimental states for everyhting here are `is this an authenticated request or not`?
No endpoint here is accessible without authentication, and all that is handled by the same auth code.

So I wanted to emphasise that at the top of the file, then with that out the way we can get down to tests (split out by endpoint) that describe the expeted functionality more.

app/lib/api_error/saved_page_not_destroyed.rb
huwd: ah that's a great point!

Initially we had these all working by rescuing raised exceptions, but most of the force pushing above is us changing that to look at validations.

Perhaps I want to change this then to `destroy`, which i assume makes the attempt, populates the errors and does not raise if something is wrong...

Then put this right and the end and say if it's still `.persisted?` then... throw this error!

huwd: or perhaps 500 and an error state is correct in this case 🤔 
I'm kind of putting the blame on the user with a 4xx error...

barrucadu: I think a 500 is the more appropriate error here, so the controller method can just be

```ruby
  def destroy
    saved_page.destroy!
    head :no_content
  end
```

And this error class doesn't need to exist

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+  end
+
+  def present_pages(saved_pages)
+    saved_pages.map { |saved_page| present_page(saved_page) }
+  end
+
+  def present_page(saved_page)
+    { "page_path" => saved_page.page_path }
+  end
```
barrucadu, line 59: Should this instead be a `to_hash` on the `SavedPage` model?

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def user
+    @user ||= OidcUser.find_by(sub: @govuk_account_session.user_id)
+  end
```
barrucadu, line 63: Use `@govuk_account_session.user` instead, which handles creating the user record if it doesn't exist.

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    if error_types.any?
+      raise ApiError::SavedPageNotPersisted, { page_path: params[:page_path] }
+    end
+  end
+
+  def present_pages(saved_pages)
+    saved_pages.map { |saved_page| present_page(saved_page) }
+  end
```
barrucadu, line 55: I'm a bit wary of single-use methods... is this really better than

```ruby
  def index
    render_api_response(saved_pages: user.saved_pages.map { |page| present_page(page) })
  end
```

Or even

```ruby
  def index
    render_api_response(saved_pages: user.saved_pages.map(&:to_hash))
  end
```

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,72 @@
+class SavedPagesController < ApplicationController
+  include AuthenticatedApiConcern
+
+  before_action :check_saved_page_exists, only: %i[destroy show]
+
+  # GET /api/saved_pages
+  def index
+    @saved_pages = SavedPage.where(oidc_user_id: user.id)
```
barrucadu, line 8: This doesn't need to be an instance variable, and also use `user.saved_pages` to make it clearer.

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+private
+
+  def handle_create_saved_page_errors(saved_page)
+    error_types = saved_page.errors.map(&:type)
+
+    if error_types.include?(:taken)
+      raise ApiError::SavedPageDuplicate, { page_path: params[:page_path] }
+    end
```
barrucadu, line 42: I'm in two minds about this.  On the one hand, saving the same page twice is a bit weird and probably shouldn't happen; but on the other hand is there any downside to making `create` do an insert-or-update?

We're not doing something like assigning each saved page a unique ID: the path is the ID.  So the API state after one POST is the same as it is after 10 POSTs.  The `create` method is idempotent.

app/lib/api_error/saved_page_not_persisted.rb
```diff
@@ -0,0 +1,21 @@
+module ApiError
+  class SavedPageNotPersisted < ApiError::Base
+    def initialize(page_path:)
+      @page_path = page_path
+    end
+
+    def status_code
+      :bad_request
```
barrucadu, line 8: Should this be a 422?

Also, this class is pretty vague.  Can the validation errors be added to the `extra_detail`?

app/models/saved_page.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def page_path_is_valid_path
+    errors.add(:page_path, :invalid_path, message: "must only include URL path and fragment") if is_invalid_path?
+  end
+
+  def is_invalid_path?
+    return true if page_path.blank?
+
+    uri = URI(CGI.unescape(page_path))
+    fragment = "##{uri.fragment}" if uri.fragment
```
barrucadu, line 23: I don't think we should have fragments, it should only be a path.

With fragments, we're implementing a "save this section of this page" feature and not a "save this page" feature.

config/locales/errors.en.yml
```diff
@@ -9,3 +9,23 @@ en:
...
+      title: Saved page not found
+      detail: A saved page with that path cannot be found associated with that user, is the path correct?
+    saved_page_not_persisted:
+      type: https://github.com/alphagov/account-api/blob/main/docs/api.md#saved_page_not_persisted
+      title: Something has gone wrong - Adding this page failed
+      detail: This page could not be persisted to the user's account. This error has been logged.
+    saved_page_not_destroyed:
+      type: https://github.com/alphagov/account-api/blob/main/docs/api.md#saved_page_not_destroyed
+      title: Something has gone wrong - Removing this page failed
+      detail: This page could not be removed from the user's account. This error has been logged.
```
barrucadu, line 23: Some of your titles start with "Something has gone wrong - ", some don't.

I think if an error is returned it's clear that something has gone wrong, so the prefix isn't needed.  Also, I've been making my titles here match the titles in the API doc.

spec/requests/saved_pages_request_spec.rb
AlanGabbianelli: I see what you mean. I added the comment to the commit where only `GET /api/saved_pages` was tested so I mistakenly thought this file was only for one endpoint, hence my suggestion.

Either
```ruby
RSpec.describe "SavedPages", type: :request do
  context "when receiving an unauthenticated request" do
    describe "GET /api/saved_pages" do
    end

    describe "GET /api/saved_pages/:page_path" do
    end
  end
    
  context "when receiving an authenticated request" do
    describe "GET /api/saved_pages" do
    end

    describe "GET /api/saved_pages/:page_path" do
    end
  end
end
```
or
```ruby
RSpec.describe "SavedPages", type: :request do
  describe "GET /api/saved_pages" do
    context "when receiving an unauthenticated request" do
    end
    
    context "when receiving an authenticated request" do
    end
  end
  
  describe "GET /api/saved_pages/:page_path" do
    context "when receiving an unauthenticated request" do
    end
    
    context "when receiving an authenticated request" do
    end
  end
end
```
works for me.

app/controllers/saved_pages_controller.rb
AlanGabbianelli: I like the `SavedPage#to_hash` suggestion. With that implemented, we can go for the nice
```ruby
def index
  render_api_response(saved_pages: user.saved_pages.map(&:to_hash))
end
```
and we don't need the two private methods.

spec/requests/saved_pages_request_spec.rb
huwd: ah ace.

cool yeah, I think if we added an endpoint that could be accessed in a logged out state (though no idea what that'd be!) I'd reorder it then but this seemed to work for now?

app/controllers/saved_pages_controller.rb
huwd: If we did though wouldn't it include other attribtues from a page?
Currently the API docs have been fairly specific about whats returned (which might be OTT?)
But so I'd wanted to constrain this and make it deliberate to add new fields so we remember to updat the API docs in the same commit?

app/controllers/saved_pages_controller.rb
huwd: nice!

app/models/saved_page.rb
huwd: Humm yeah I see what you mean, though we will control which link we hand to the button to persist right?

Mostly what I was thinking about with this was the `/bank-holiday#scotland` URL, where a user may want to save that specific page, but to lose the framgent gets them to different content?

app/lib/api_error/saved_page_not_persisted.rb
huwd: Yeah it is pretty vauge, it fell into a similar category as the other destroy method.
This one was initially what was returned once as a final else to a `rescue` if we'd already rescued all known errors (ones with specific messages), and yet the resource remained not `.persisted?`

However since switching these to use ActiveModel validations it gets a bit more compliacted.
We want to use `save` on the model, as we can then capture `:blank` and return a more meaingful error name (if we throw an error I need to return to using rescue as flow control).

However there may be errors that I have not handled for, in that case I'd wanted to a generic error that I throw as a last ditch effort. Again so if these pop up in sentry we can act on them quickly, but also to cater for any scenario i've not thought about.

I considered if these were 5xx errors (our fault) or 4xx (theirs), and went with bad request because I reckon to end up here there must have been a wacky request that:
- was authenticated
- had a non blank, valid page path

But yet could not be persisted...

config/locales/errors.en.yml
huwd: These are fairly vague because they're fallbacks, I can't quite imagine how folks could end up on the branch that raises this error, but I don't want to leave those routes unguarded.

The destroy I think you are right about above so I'll replace that with a 5xx error state.
But I think because we're not doing `save!` I can't use the same on the `save_page_not_persisted`.

It's vague becuase if we ever saw it we'd inevitably want to go back and write a more specific error?

app/controllers/saved_pages_controller.rb
barrucadu: > If we did though wouldn't it include other attribtues from a page?

Depends on what we make `to_hash` return

app/models/saved_page.rb
barrucadu: I think it's a lesser can of worms to handle the bank holidays page with the user's saved location, and to keep this feature for saving pages (and not parts of pages)

app/lib/api_error/saved_page_not_persisted.rb
barrucadu: Ok, in that case, rather than having a separate errors for some validation errors, why not have one error type for all validation errors?

```ruby
raise ApiError::CannotSavePage, { page_path: params[:page_path], errors: saved_page.errors }
```

I think a useful thing to keep in mind for introducing new API errors is whether splitting an error into more specific types lets the API client react in a way it didn't have enough information for before.  Is an API client really going to handle `SavedPageDuplicate`, `SavedPagePathRequired`, and `SavedPageNotPersisted` differently?

AlanGabbianelli: Probably not, that's a good point. So `create` would just be
```ruby
  # POST /api/saved_pages
  def create
    saved_page = SavedPage.new(oidc_user_id: user.id, page_path: params[:page_path])

    if saved_page.save
      render_api_response(saved_page: saved_page.to_hash)
    else
      raise ApiError::CannotSavePage, { page_path: params[:page_path], errors: saved_page.errors }
    end
  end
```

app/controllers/saved_pages_controller.rb
```diff
@@ -1,14 +1,19 @@
...
 
   # GET /api/saved_pages
   def index
     @saved_pages = SavedPage.where(oidc_user_id: user.id)
     render_api_response(saved_pages: present_pages(@saved_pages))
   end
 
+  # GET /api/saved_pages/:page_path
+  def show
+    render_api_response(saved_page: present_page(saved_page))
```
AlanGabbianelli, line 14: If you implement the `SavedPage#to_hash` method, this can become
```ruby
render_api_response(saved_page: saved_page.to_hash)
```

app/controllers/saved_pages_controller.rb
AlanGabbianelli: Based on the `present_page(saved_page)` method, `SavedPage#to_hash` can be

```ruby
def to_hash
  { "page_path" => page_path }
end
```

huwd: humm, `to_hash` might confuse with the `to_h` method that would do something similar but more potentially?
I went with `present` as `presenters` are an established pattern for formatting and filtering return values I'd thought?

app/models/saved_page.rb
huwd: Ok cool I'll scrub that bit

app/controllers/saved_pages_controller.rb
huwd: much better thank you, have resolved and will force push once I've tidied up a few other bits.

app/controllers/saved_pages_controller.rb
huwd: Done, great improvement thanks!

app/controllers/saved_pages_controller.rb
AlanGabbianelli: We already use `to_hash` in a few places ([here in account-api](https://github.com/alphagov/account-api/blob/056066400e0edd8be98fbf65cc76159027d242d5/app/lib/account_session.rb#L68-L75) and [here in govuk-account-manager-prototype](https://github.com/alphagov/govuk-account-manager-prototype/blob/7858b55afcddc2f855d909abd051838d95f21f13/app/models/ephemeral_state.rb#L7-L13)) so it would be consistent with what we do already. But I don't see anything wrong with `present` either.

barrucadu: I think I'd prefer something based on "presenter" than just "present", as there's already a `present?` method

app/controllers/saved_pages_controller.rb
huwd: Fair point, it can be shrunk, let's roll that conversation into this one: https://github.com/alphagov/account-api/pull/69/files#r633509206

app/controllers/saved_pages_controller.rb
huwd: Ok let's roll that conversation into this one: https://github.com/alphagov/account-api/pull/69/files#r633509206

app/controllers/saved_pages_controller.rb
huwd: Yeah I see what you mean, Initially i'd wanted this for the alert we could send the user, essentially:
> your trying to save something that is already saved, don't worry it's we've got the thing you want on file, here it is.

just saving it and showing the success state achieves the same thing but cleaner...
🤔 but I'm wondering about the code we'll inevitably need to add for what happens if guidance is `gone` or `[withdrawn]` or redirected.

Updates there might mean they go to create something then find out its a duplicate later?
Wouldn't a specific message that highlights that be useful there? If so, this pattern solves a fairly rare problem now and might make that easier later?

app/lib/api_error/saved_page_not_persisted.rb
huwd: Humm, One unified error message sounds good, but I'm also a bit nervous of sending raw errors back out over HTTP?

Stop me if htis is too paranoid?
But i never quite know the full range of what we might expose there, wouldn't want some innards of the app to give any good intellegence to an attacker? Or at least one who hadn't found the source code here yet...

huwd: I guess what'd i'd been trying to do is capture specific error states where I thought "there's a likelyhood this could crop up" and return specific, actionable error messages.

Then at the bottom have an `else` category, where something has gone wrong, but I didn't forsee what it was here.
If we ever saw that in sentry it'd be a guide for devs to go back, investigate and likely add another test case + more specific error, or fix a bug

barrucadu: We're not giving away any internal details of the app because we're not directly exposing account-api publicly.  And also returning validation errors in a 422 response is a pretty common pattern for our other API apps (eg, publishing-api does it).

app/controllers/saved_pages_controller.rb
barrucadu: > your trying to save something that is already saved, don't worry it's we've got the thing you want on file, here it is.

Is that useful and actionable information for the user?  What will someone do differently in the case that a page is already saved?

> Updates there might mean they go to create something then find out its a duplicate later?

If the only route to save a page is via a button on the page which shows a save/unsave state based on the `show` response, then the only way a user could save a page twice is if they had two tabs with the same page open, or if there were a bug in the button logic.

If, for example, a page gets redirected and we update the path in the database automatically, then when the user looks at the new page the button will be in the saved state.

huwd: Ok so then it's back to a POST updating things 🤔 
So to run down this, to impliment this I think i'd be looking at:
- removing the unique index on the table
- removing the duplicate validation
- changing the post controller to use find_or_create_by
- removing the error, and the docs

If we're on the fence about it, then better to leave it?
If we've swung towards removing it in this thread then I'll go remove those parts?

barrucadu: > removing the unique index on the table
> removing the duplicate validation

We definitely still want the uniqueness, to protect against application bugs.  But I don't think we want to treat it as an API error to save the same page twice.

app/models/saved_page.rb
```diff
@@ -1,9 +1,37 @@
...
+
+  def page_path_is_valid_path
+    errors.add(:page_path, :invalid_path, message: "must only include URL path") if is_invalid_path?
+  end
+
+  def is_invalid_path?
+    return true if page_path.blank?
+
+    uri = URI(CGI.unescape(page_path))
+    page_path != CGI.escape(uri.path)
```
barrucadu, line 29: Why are we working with escaped strings?

If the request is something like

```
GET /saved-pages/%2Fsome%2Fpage%2Fon%2Fgov%2Fuk
```

Doesn't Rails automatically escape that for us in the controller to

```
params[:page_path] = "/some/page/on/gov/uk`
```

?

Manually (un)escaping strings is a place where errors could easily be introduced.

app/models/saved_page.rb
```diff
@@ -1,9 +1,37 @@
...
+  end
+
+  def is_url_unsafe_string?
+    return true if page_path.nil?
```
AlanGabbianelli, line 33: Should this be `return true if page_path.blank?` as in `is_invalid_path?`?

app/models/saved_page.rb
barrucadu: Ah, I also see `is_url_unsafe_string?`

Right, we should be storing actual paths in our database, not escaped versions of them.

docs/api.md
```diff
@@ -558,6 +559,58 @@ Response:
...
+```
+
+Response when a user has saved two pages:
+
+```json
+{
+    "govuk_account_session": "YWNjZXNzLXRva2Vu.cmVmcmVzaC10b2tlbg==",
+    "saved_pages": [
+      { "page_path": "%2Fgovernment%2Forganisations%2Fgovernment-digital-service" },
+      { "page_path": "%2Fgovernment%2Forganisations%2Fcabinet-office" },
```
barrucadu, line 614: As mentioned in the previous commit, we should be working with actual paths, not escaped ones.  This is introducing potential points of confusion and bugs.

For example, the attribute service prototype expects attribute values to be passed around as encoded JSON strings.  We had an issue where [those strings weren't being unencoded](https://github.com/alphagov/govuk-attribute-service-prototype/pull/112), and so new users who signed up weren't able to return to their saved results.  If we'd just never had the additional encoding step, that problem wouldn't have occurred.

config/locales/errors.en.yml
```diff
@@ -9,3 +9,7 @@ en:
       type: https://github.com/alphagov/account-api/blob/main/docs/api.md#unknown-attribute-names
       title: Unknown attribute names
       detail: Attribute names %{attributes} are unknown, have they been added to config/user_attributes.yml?
+    saved_page_not_found:
+      type: https://github.com/alphagov/account-api/blob/main/docs/api.md#saved_page_not_found
+      title: Saved page not found
+      detail: A saved page with that path cannot be found associated with that user, is the path correct?
```
barrucadu, line 15: This should be the `cannot_save_page` error

barrucadu: Also, for the other errors, the class name and i18n keys match the title, so it's dead simple to see what is related to what: `LevelOfAuthenticationTooLow => level_of_authentication_too_low => "Level of authentication too low"`

docs/api.md
```diff
@@ -611,6 +613,56 @@ Response when a user has saved two pages:
...
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.set_saved_page(
+    page_path: "/guidance/foo",
+    govuk_account_session: "session-identifier",
+
```
barrucadu, line 657: There's an extra newline here

spec/models/saved_page_spec.rb
```diff
@@ -1,4 +1,55 @@
...
+
+    it "allows saving URL safe strings" do
+      saved_page.page_path = CGI.escape("/foo/bar")
+      expect(saved_page).to be_valid
+    end
+
+    it "does not allow empty strings as paths" do
+      saved_page.page_path = ""
+      expect(saved_page).not_to be_valid
+    end
```
AlanGabbianelli, line 38: Nitpick, feel free to ignore.
For consistency with other tests, this should come right after the `"rejects nil"` test, have title `"reject empty strings"` and have the error message expectation.

AlanGabbianelli: Or the two could be tested together (since it's the same validation):
```ruby
it "rejects blank values as page paths" do
  [nil, ""].each do |blank_value|
    saved_page.page_path = blank_value
    expect(saved_page).not_to be_valid
    expect(saved_page.errors[:page_path]).to include("can't be blank")
  end
end
```

docs/api.md
```diff
@@ -664,6 +666,40 @@ GdsApi.saved_page_api.set_saved_page(
...
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.delete_saved_page.get_saved_page(
+    page_path: "/guidance/bar",
+    govuk_account_session: "session-identifier",
+
```
barrucadu, line 698: Extra newline here

docs/api.md
```diff
@@ -26,6 +26,7 @@ management. This API is not for other government services.
   - [`GET /api/saved-pages`](#get-apisaved-pages)
   - [`POST /api/saved-pages`](#post-apisaved-pages)
   - [`DELETE /api/saved-pages/:page_path`](#delete-apisaved-pagespage_path)
+  - [`GET /api/saved-pages/:page_path`](#get-apisaved-pagespage_path)
```
barrucadu: These are all at `/api/saved_pages`, not `/api/saved-pages`, right?  I think we'd need to change the routes file to get a `-`

docs/api.md
```diff
@@ -700,6 +701,54 @@ GdsApi.delete_saved_page.get_saved_page(
...
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.get_saved_page(
+    page_path: "/guidance/bar",
+    govuk_account_session: "session-identifier",
+
```
barrucadu, line 739: Extra newline

config/locales/errors.en.yml
barrucadu: Ok, now looking at the full diff, the problem is that these `saved_page_not_found` locale entries are introduced in the "create a saved page" commit, and the locale entries for the `CannotSavePage` class are not added anywhere.

app/models/saved_page.rb
```diff
@@ -0,0 +1,26 @@
...
+private
+
+  def page_path_is_valid_path
+    errors.add(:page_path, :invalid_path, message: "must only include URL path") if is_invalid_path?
+  end
+
+  def is_invalid_path?
+    return true if page_path.blank?
+
+    uri = URI(page_path)
```
barrucadu, line 23: Can we make this a bit more robust to bad input?  Currently it'll throw an exception if given something that doesn't look like a URI:

```
irb(main):010:0> URI("hello world")
Traceback (most recent call last):
        1: from (irb):10
URI::InvalidURIError (bad URI(is not URI?): "hello world")

irb(main):011:0> URI(nil)
Traceback (most recent call last):
        2: from (irb):10
        1: from (irb):11:in `rescue in irb_binding'
ArgumentError (bad argument (expected URI object or URI string))
```

huwd: That's true, however I think blank / empty is the main case here and that guard clause on L21 covers that. Plus they'd get a warning about `can't be blank` from the rails validations too.

I've tested a bunch of these and URI seems pretty open:

```
URI("https://gov.uk/foo")                                                                                                                                                                         
=> #<URI::HTTPS https://gov.uk/foo>

URI("www.gov.uk/foo")                                                                                                                                                                             
=> #<URI::Generic www.gov.uk/foo>

URI("gov.uk/foo")                                                                                                                                                                                 
=> #<URI::Generic gov.uk/foo>

URI("/foo")                                                                                                                                                                                       
=> #<URI::Generic /foo>

URI("foo")                                                                                                                                                                                        
=> #<URI::Generic foo>
```

If it is none of those and not nil, then I kind of think throwing an error might be good!

huwd: It actually also works with 
```
URI("")                                                                                                                                                                                           
=> #<URI::Generic >
```

... which tbh seems pretty bad

barrucadu: > If it is none of those and not nil, then I kind of think throwing an error might be good!

It should be a validation error we return to the user, not a 5xx error due to our validation code throwing an uncaught exception

huwd: I looked into other ways of doing it, URI comes with some regex that you can use for comparisons.
However here we don't really care about the rest of the URI. it's just if it can be parsed into URI.path and if the output matches

barrucadu: How about

```ruby
def is_valid_path?(page_path)
  URI(page_path) == page_path
rescue StandardError
  false
end
```

Then change line 17 to use `unless is_valid_path?`

huwd: oh actually turns out `URI("foo").path` becomes `foo` so I might just add .starts_with("/") as a condition too

config/locales/errors.en.yml
huwd: rebase-tastic... Sorry this thing has been rebased about 💯  x 💯 times now, it's easy for things to get lost along the way.
I'll get that back.

config/locales/errors.en.yml
```diff
@@ -9,3 +9,11 @@ en:
       type: https://github.com/alphagov/account-api/blob/main/docs/api.md#unknown-attribute-names
       title: Unknown attribute names
       detail: Attribute names %{attributes} are unknown, have they been added to config/user_attributes.yml?
+    saved_page_not_found:
+      type: https://github.com/alphagov/account-api/blob/main/docs/api.md#saved_page_not_found
```
barrucadu, line 13: The anchors which GitHub generates use `-`s not `_`s

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
```
barrucadu, line 631: This is a path parameter, not a request parameter

barrucadu: Or maybe a "path request parameter"

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+- `saved_page`
+  - an object containing the page path of the successfully saved page
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 422 if page path is nil or empty
+- 422 if page path is not an escaped URL safe string
+- 422 if page path is not just a path
+- 422 if the page could not otherwise be successfully saved
```
barrucadu, line 646: Collapse all these into one: `422 if the page could not be saved (see [error: page cannot be saved](#page-cannot-be-saved))`

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+- 200 otherwise
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.set_saved_page(
```
barrucadu, line 654: Why `set_saved_page` and not `save_page`?

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    @user ||= @govuk_account_session.user
+  end
+
+  def saved_page
+    @saved_page ||= SavedPage.find_by(oidc_user_id: user.id, page_path: params[:page_path])
+  end
+
+  def check_saved_page_exists
+    raise ApiError::SavedPageNotFound, { page_path: params[:page_path] } if saved_page.nil?
+  end
```
barrucadu, line 45: A thought: is this new error type actually valuable?  What does it let a client app do that a regular 404 does not?

- `UnknownAttributeNames` gives a list of the unknown attributes, which is useful, as the client can request several attributes at once and only some may be invalid.  That's not the case here, as the client only requests a single saved page.
- `LevelOfAuthenticationTooLow` returns the level that the client needs to reauthenticate the user at, otherwise the client would have to guess.  That's not the case here, as the error doesn't explain how to fix the problem.
- `CannotSavePage` returns the list of validation errors, which could be logged or shown to the user, letting them fix the problem themselves (though, probably logged rather than shown).  Here there's just one error: this page hasn't been saved.

So, I'm leaning towards this being just a plain body-less 404 response.

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
```
barrucadu, line 681: This is also a path param

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+- 401 if the session identifier is invalid
+- 404 cannot find a page with the provided path
+- 204 successfully deleted
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.delete_saved_page.get_saved_page(
```
barrucadu, line 694: This should be something like `GdsApi.saved_page_api.delete_saved_page`, right?

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 404 cannot find a page with the provided path
```
barrucadu, line 686: If you do keep the separate 404 error type, then add a link to its docs here.

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.delete_saved_page.get_saved_page(
+    page_path: "/guidance/bar",
+    govuk_account_session: "session-identifier",
+)
+```
+
+Response is status code only unless there are error details (see below)
```
barrucadu, line 700: List the status codes

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Request parameters
+
+- `page_path`
+  - An escaped URL safe string that identifies the path of a saved page.
```
barrucadu, line 714: Path parameter

docs/api.md
```diff
@@ -554,6 +564,187 @@ Response:
...
+
+- `govuk_account_session` *(optional)*
+  - a new session identifier
+- `saved_page`
+  - an object containing the page path of the successfully queried page
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 404 cannot find a page with the provided path
```
barrucadu, line 726: Here too, if you keep the separate 404 type, link to the doc

docs/api.md
```diff
@@ -589,8 +780,25 @@ level to access.
...
+
 #### Debugging steps
 
 - check that you don't have a typo in the attribute names
 - check that the attributes are defined in `config/user_attributes.yml`
 - check that you are running the latest version of account-api
+
+### Page cannot be saved
+
+An attempt was made to save the page with the provided page path, however it could not be persisted.
```
barrucadu, line 799: "An attempt was made" is redundant, as all the endpoints attempt to do the thing they should.  Also this should mention the `errors` field.  How about

> The page you have specified could not be saved.  The `errors` response field lists the problems.

docs/api.md
```diff
@@ -589,8 +780,25 @@ level to access.
...
+
+#### Debugging steps
+
+- check the `errors` returned as an extra detail in the response for specific error messages
+- ensure the page_page path does not violate any database constraints or validations
```
barrucadu, line 804: Users of the API don't (and shouldn't) need to know about our internal database validations.  We should provide a well-specified API they can use without digging into the code.

Since the `errors` field includes the problems, I think this "database constraints or validations" line can be dropped.

spec/models/saved_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      expect(saved_page.errors[:page_path]).to include("can't be blank")
+    end
+
+    it "rejects blank values as page paths" do
+      [nil, ""].each do |blank_value|
+        saved_page.page_path = blank_value
+        expect(saved_page).not_to be_valid
+        expect(saved_page.errors[:page_path]).to include("can't be blank")
+      end
+    end
```
barrucadu, line 30: This second test covers the first one

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+      it "returns an empty array if there are no saved pages" do
+        get saved_pages_path, headers: headers
+
+        expect(response).to have_http_status(:ok)
+        expect(JSON.parse(response.body)["saved_pages"]).to eq([])
+      end
+
+      it "returns an array of saved_pages if they exist" do
+        list = FactoryBot.create_list(:saved_page, 2, oidc_user_id: user.id)
+        expected_response = list.map { |page| { "page_path" => page.page_path } }
```
barrucadu, line 51: `expected_response = list.map(&:to_hash)`

app/controllers/saved_pages_controller.rb
barrucadu: This is an optional suggestion

app/controllers/saved_pages_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    saved_page = SavedPage.find_or_create_by(oidc_user_id: user.id, page_path: params[:page_path])
+
+    if saved_page.persisted?
+      render_api_response(saved_page: saved_page.to_hash)
+    else
+      raise ApiError::CannotSavePage, { page_path: params[:page_path], errors: saved_page.errors }
+    end
+  end
+
+  # DELETE /api/saved_pages
```
barrucadu, line 27: The route is `DELETE /api/saved_pages/:page_path` according to `rails routes`

docs/api.md
barrucadu: Whoops, no, this *is* a request parameter

docs/api.md
barrucadu: Hmm, I guess that's implicit from the URL in the section title actually... so this is fine

docs/api.md
barrucadu: This is also fine

barrucadu:
When this PR is good to go, before merging it it would be best to make sure the pact tests in your gds-api-adapters branch pass against it, so we don't immediately need to open another account-api PR to fix any issues

spec/requests/saved_pages_request_spec.rb
AlanGabbianelli: That would be calling the same code that we are testing. How about something like this?
```ruby
it "returns an array of saved_pages if they exist" do
  page1, page2 = FactoryBot.create_list(:saved_page, 2, oidc_user_id: user.id)
  expected_response = [
    { "page_path" => page1.page_path },
    { "page_path" => page2.page_path },
  ]

  get saved_pages_path, headers: headers

  expect(response).to have_http_status(:ok)
  expect(JSON.parse(response.body)["saved_pages"]).to match_array(expected_response)
end
```
(`match_array` instead of `eq` because we don't care about the order)

docs/api.md
huwd: I was following what I thought was a convention in gds api adaptors from account api:
https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/account_api.rb#L48-L65

Everything else seemed to be `get_`, `set_` or `check_`

But not wedded to it at all, happy to change.

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+        expect(JSON.parse(response.body)["saved_page"]).to eq({ "page_path" => "/page-path/1" })
+      end
+
+      it "returns status 200 and upserts the record if the page already exists" do
+        FactoryBot.create(:saved_page, oidc_user_id: user.id, page_path: "/page-path/1")
+        post saved_pages_path, headers: headers, params: { page_path: "/page-path/1" }.to_json
+
+        expect(response).to have_http_status(:ok)
+        expect(JSON.parse(response.body)["saved_page"]).to eq({ "page_path" => "/page-path/1" })
+      end
```
AlanGabbianelli, line 74: We're not testing that a record gets created in the first test and that it doesn't in the second test.

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    describe "DELETE /api/saved_pages" do
+      it "returns status 204 No Content if a record has been sucessfully deleted" do
+        FactoryBot.create(:saved_page, oidc_user_id: user.id, page_path: "/page-path/1")
+        delete saved_page_path(page_path: "/page-path/1"), headers: headers
+
+        expect(response).to have_http_status(:no_content)
+      end
```
AlanGabbianelli, line 133: We're not testing that a record gets deleted.

docs/api.md
```diff
@@ -558,6 +559,57 @@ Response:
...
+
+- 401 if the session identifier is invalid
+- 200 otherwise
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.saved_page_api.get_saved_pages(
```
huwd, line 592: if the `get_`, `set_`, `check_` isn't a convention then this could just be `.saved_pages`.

But perhaps we can see what names come out of the review of GDS API adaptors and I can check the docs match the reality once those are merged?

barrucadu: I think the closest similar thing is the publishing-api's put content endpoint: https://github.com/alphagov/publishing-api/blob/main/docs/api.md#put-v2contentcontent_id

Which uses a PUT rather than a POST and has the ID as a path param, which makes sense as the ID is known without needing to talk to publishing-api first.  So I'd be in favour of making our save-a-new-page endpoint like that.

barrucadu: For getting all content, the publishing-api uses [get_content_items](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api.rb#L352), so I think `get_saved_pages` is fine.

spec/requests/saved_pages_request_spec.rb
huwd: [Done](https://github.com/alphagov/account-api/pull/69/commits/0c4ad32ab6c9b7f9d1be36f59645c6689a838a7b#diff-e494be0ec09eb92d5782bcb4d976f0a1c24e9f3874d5b2a1e980ac902d51eda7R56-R60)

spec/requests/saved_pages_request_spec.rb
huwd: [Done](https://github.com/alphagov/account-api/pull/69/commits/ed5f50b71aa55fa860315f20d0802db74e244b34#diff-e494be0ec09eb92d5782bcb4d976f0a1c24e9f3874d5b2a1e980ac902d51eda7R127-R133)

docs/api.md
```diff
@@ -554,6 +563,184 @@ Response:
...
+- 401 if the session identifier is invalid
+- 404 cannot find a page with the provided path
+- 204 successfully deleted
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.delete_saved_page.delete_saved_page(
```
barrucadu, line 690: This says `GdsApi.delete_saved_page.delete_saved_page`, but should be `GdsApi. saved_page_api.delete_saved_page` - though, see my comment in gds-api-adapters about the separate class

huwd: Done - yeah that was just copy pasta, It's a bit strange doing this bit of the docs, perhaps this should have been a follow up commit.

I figured I'd do the GDS API adaptors then come back and check these matched. For all I know we could rename a bunch of those in the review over there

spec/requests/saved_pages_request_spec.rb
```diff
@@ -0,0 +1,148 @@
+RSpec.describe "SavedPages", type: :request do
```
barrucadu, line 1: Based on [the rspec docs](https://relishapp.com/rspec/rspec-rails/v/3-8/docs/request-specs/request-spec) I think it's more conventional for this file to be `spec/requests/saved_pages_spec.rb` and for this first line to be

```ruby
RSpec.describe "Saved pages" do
```

spec/service_consumers/pact_helper.rb
```diff
@@ -106,6 +106,53 @@ def url_encode(str)
     end
   end
 
+  provider_state "there is a valid user session, with saved pages" do
+    set_up do
```
huwd, line 110: I've had issues with these, its seems like they don't clear up behind themselves in the same way as before / after blocks.

So i've had state bleeding between my tests.
I've tried this and it works but I really don't like it. It feels messy and horrible

spec/service_consumers/pact_helper.rb
```diff
@@ -106,6 +106,53 @@ def url_encode(str)
...
+      SavedPage.destroy_all
+      user = OidcUser.find_or_create_by(sub: "user-id")
```
huwd, line 112: this feels particularly wrong but I'm getting:
```
     ActiveRecord::RecordNotUnique:
       PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_oidc_users_on_sub"
       DETAIL:  Key (sub)=(user-id) already exists.
```

if i switch it to:
```
      user = FactoryBot.create(:oidc_user, sub: "user-id")
```

barrucadu: That's because `FactoryBot.create(:oidc_user, sub: "user-id")` is trying to recreate the user record which may already exist.

spec/service_consumers/pact_helper.rb
```diff
@@ -106,6 +106,53 @@ def url_encode(str)
...
+
+  provider_state "there is a valid user session, a request to save /guidance/some-govuk-guidance saved" do
+    set_up do
+      OidcUser.find_or_create_by!(sub: "user-id")
+    end
+
+    tear_down do
+      OidcUser.find_by(sub: "user-id").destroy!
+    end
+  end
```
barrucadu, line 142: Why is this needed over just `there is a valid user session`?

spec/service_consumers/pact_helper.rb
```diff
@@ -106,6 +106,53 @@ def url_encode(str)
...
+      user = OidcUser.find_or_create_by!(sub: "user-id")
+      FactoryBot.create(:saved_page, page_path: "/guidance/some-govuk-guidance", oidc_user_id: user.id)
+    end
+
+    tear_down do
+      OidcUser.find_by(sub: "user-id").destroy!
+      SavedPage.destroy_all
+    end
+  end
+
```
barrucadu, line 155: Why is this needed over just `there is a valid user session, with /guidance/some-govuk-guidance saved`?

spec/service_consumers/pact_helper.rb
barrucadu: Why is this `SavedPage.destroy_all` needed?

huwd: But shouldn't the test database be empty at this point?
I don't see why that record should already exist?

spec/service_consumers/pact_helper.rb
huwd: humm there might be a clash no?
I could put either delete or save into the "there is a valid session",
But the happy path for each depends on opposite state (delete is happy if there was something to delete, create is happy if there was nothing to create?)

spec/service_consumers/pact_helper.rb
barrucadu: I don't think it runs each test in a transaction

spec/service_consumers/pact_helper.rb
barrucadu: A clash of what?  Right now this `provider_state` isn't setting anything up relating to saved pages

spec/service_consumers/pact_helper.rb
huwd: what do you think the best approach is? all these setup and teardown blocks seem a bit lame?

I could create a user once at the top perhaps?

barrucadu: > I could create a user once at the top perhaps?

Yes, my fixup commit does that, and makes all the tests use it instead of specifying `user-id` everywhere

spec/requests/saved_pages_request_spec.rb
barrucadu: I've unresolved this because it's not been resolved

PR #72, opened 2021-05-19
Set the updated_at time when storing new local attribute values

barrucadu:
By default upsert_all does not change the updated_at timestamp.  This
is fine if nothing has actually changed (eg, saving a duplicate page);
but in general saving an attribute value *is* a change, and so we want
the timestamp to change.

### DIFF
--- a/spec/lib/account_session_spec.rb
+++ b/spec/lib/account_session_spec.rb
@@ -148,6 +148,18 @@ RSpec.describe AccountSession do
         expect(stub).to have_been_made
       end
 
+      context "when the local attribute already exists" do
+        it "increases the updated_at time" do
+          attribute = LocalAttribute.create!(
+            oidc_user: OidcUser.find_or_create_by(sub: user_id),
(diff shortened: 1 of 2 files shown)


spec/lib/account_session_spec.rb
```diff
@@ -148,6 +148,20 @@
...
+        it "increases the updated_at time" do
+          attribute = LocalAttribute.create!(
+            oidc_user: OidcUser.find_or_create_by(sub: user_id),
+            name: local_attribute_name,
+            value: local_attribute_value,
+          )
+
+          updated_at = attribute.updated_at
+          account_session.set_attributes(local_attributes)
+          expect(attribute.reload.updated_at).not_to eq(updated_at)
```
AlanGabbianelli, line 161:
```suggestion
          expect { account_session.set_attributes(local_attributes) }.to change { attribute.reload.updated_at }
```

AlanGabbianelli (approved):

PR #75, opened 2021-05-21
Implement the message queue processor

barrucadu:
For "gone" notifications, redirect if there is an alternative path or
delete the saved page if not.

For "redirect" notifications, redirect if there is a matching
redirect (including gracefully handling prefix redirects) or delete
the saved page if not.

For "vanish" notifications, delete the saved page.

For other notifications, update the saved information.

See also https://github.com/alphagov/govuk-puppet/pull/11111

---

[Trello card](https://trello.com/c/8EadgviV/772-get-account-api-consuming-publishing-messages-from-rabbitmq)

### DIFF
--- a/lib/message_queue_processor.rb
+++ b/lib/message_queue_processor.rb
@@ -1,0 +1,92 @@
+class MessageQueueProcessor
+  def process(message)
+    summary = process_message(message.payload)
+    message.ack
+    Rails.logger.info summary.to_json
+  end
+
+  # For the details of unpublishing types, the publishing-api docs are
+  # the canonical source:
+  # https://github.com/alphagov/publishing-api/blob/main/docs/model.md#unpublishing
+  def process_message(payload)
+    affected = SavedPage.where(content_id: payload["content_id"])
+    affected_count = affected.count
+
+    effect =
+      case payload["document_type"]
+      when "gone"
+        redirect_saved_pages(affected, find_alternative_path(payload))
+      when "redirect"
+        redirect_saved_pages(affected, find_alternative_path(payload))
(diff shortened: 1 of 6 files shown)


lib/message_queue_processor.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def find_alternative_path(payload)
+    alternative_path = payload.dig("details", "alternative_path")
+    return alternative_path if alternative_path
+
+    exact_redirect = payload["redirects"]&.find { |r| r["type"] == "exact" && payload["base_path"] == r["path"] }
+    return exact_redirect["destination"] if exact_redirect
+
+    prefix_redirect = payload["redirects"]&.select { |r| r["type"] == "prefix" && payload["base_path"].start_with?(r["path"]) }&.max_by { |r| r["path"].length }
```
huwd, line 43: Interested so out of a number of possible redirects here, we're filtering then ultimately selecting the longest by path?
Why the longest?

barrucadu: That's the behaviour router uses.  Say you visit `/government/organisations/ministry-of-magic/spellcasting-policy`, it'll look for these routes in this order:

1. An exact route for `/government/organisations/ministry-of-magic/spellcasting-policy`
2. A prefix route for `/government/organisations/ministry-of-magic/spellcasting-policy`
3. A prefix route for `/government/organisations/ministry-of-magic`
4. A prefix route for `/government/organisations`

huwd: makes sense, cheers for the explaination!

huwd (approved):

PR #78, opened 2021-05-24
Cache the OIDC discovery response

barrucadu:
We use OIDC discovery to avoid hard-coding things like URLs or keys in
this app.  This method makes an HTTP request to the Identity
Provider (govuk-account-manager-prototype) and then validates the
response.  But this response changes very infrequently, so this is
adding an unnecessary (and slow) network hop to almost every
account-api response.

This is most visible in the auth endpoint which, despite only
generating a URL, commonly takes 150ms because it's got this extra
request and validation going on, but it impacts the attribute and
Transition Checker email endpoints too.

---

[Trello card](https://trello.com/c/F4q6raJf/785-cache-oidc-discovery-response)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ ruby File.read(".ruby-version").strip
 gem "rails", "6.1.3.2"
 
 gem "bootsnap"
+gem "dalli"
 gem "gds-sso"
 gem "govuk_app_config"
 gem "openid_connect"
--- a/spec/support/helpers/oidc_client_helper.rb
+++ b/spec/support/helpers/oidc_client_helper.rb
@@ -1,15 +1,14 @@
 module OidcClientHelper
   def stub_oidc_discovery
-    discovery_response = instance_double(
-      "OpenIDConnect::Discovery::Provider::Config::Response",
+    discovery_response = {
       authorization_endpoint: "http://openid-provider/authorization-endpoint",
       token_endpoint: "http://openid-provider/token-endpoint",
       userinfo_endpoint: "http://openid-provider/userinfo-endpoint",
       end_session_endpoint: "http://openid-provider/end-session-endpoint",
-    )
+    }
 
     # rubocop:disable RSpec/AnyInstance
-    allow_any_instance_of(OidcClient).to receive(:discover).and_return(discovery_response)
+    allow_any_instance_of(OidcClient).to receive(:cached_discover_response).and_return(discovery_response)
     # rubocop:enable RSpec/AnyInstance
   end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -90,6 +90,7 @@ GEM
     crack (0.4.5)
       rexml
     crass (1.0.6)
+    dalli (2.7.11)
     diff-lcs (1.4.4)
     dig_rb (1.0.1)
     docile (1.3.5)
@@ -416,6 +417,7 @@ DEPENDENCIES
   bootsnap
   bullet
   climate_control
+  dalli
   factory_bot_rails
   gds-sso
   govuk_app_config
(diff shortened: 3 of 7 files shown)


spec/support/helpers/oidc_client_helper.rb
```diff
@@ -1,15 +1,14 @@
...
       token_endpoint: "http://openid-provider/token-endpoint",
       userinfo_endpoint: "http://openid-provider/userinfo-endpoint",
       end_session_endpoint: "http://openid-provider/end-session-endpoint",
-    )
+    }
 
     # rubocop:disable RSpec/AnyInstance
-    allow_any_instance_of(OidcClient).to receive(:discover).and_return(discovery_response)
+    allow_any_instance_of(OidcClient).to receive(:cached_discover_response).and_return(discovery_response)
     # rubocop:enable RSpec/AnyInstance
```
AlanGabbianelli, line 12: I know it wasn't introduced in this PR but to avoid disabling this cop, we can do this:
```ruby
oidc_client = instance_double(OidcClient)
allow(OidcClient).to recive(:new).and_return(oidc_client)
allow(oidc_client).to receive(:cached_discover_response).and_return(discovery_response)
```
I haven't double-checked the syntax but it should be pretty close.

AlanGabbianelli: Or we can just disable the cop in `.rubocop.yml` or `rubocop-govuk` since we seem to use `allow_any_instance_of` quite a bit: https://github.com/search?q=org%3Aalphagov+allow_any_instance_of&type=code
(out of scope here, I know 😃)

barrucadu: Using a double doesn't work here because we want all the *other* methods to behave as they currently do.  We'd need a double which only overrides one method and delegates all the others to a real instance:

```
  1) Transition Checker email subscriptions GET when the user has a deactivated email subscription returns 'false'                                                                                           
     Failure/Error:                                                                                                                                                                                          
       oauth_response = oidc_client.public_send(                                                                                                                                                             
         method,                                                                                                                                                                                             
         **args.merge(access_token: @access_token, refresh_token: @refresh_token),                                                                                                                           
       )                                                                                                                                                                                                     
                                                                                                                                                                                                             
       #<InstanceDouble(OidcClient) (anonymous)> received unexpected message :has_email_subscription with ({:access_token=>"access-token", :refresh_token=>"refresh-token"})                                 
     # ./app/lib/account_session.rb:131:in `public_send'                                                                                                                                                     
     # ./app/lib/account_session.rb:131:in `oidc_do'                                                                                                                                                         
     # ./app/lib/account_session.rb:93:in `has_email_subscription?'                                                                                                                                          
     # ./app/controllers/transition_checker_email_subscription_controller.rb:7:in `show'                                                                                                                     
     # ./spec/requests/transition_checker_email_subscription_spec.rb:34:in `block (4 levels) in <top (required)>'                                                                                            
     # ./spec/spec_helper.rb:33:in `block (3 levels) in <top (required)>'                                                                                                                                    
     # ./spec/spec_helper.rb:32:in `block (2 levels) in <top (required)>'                    

(+ 50 more failures)
```

Gemfile
```diff
@@ -8,6 +8,7 @@ ruby File.read(".ruby-version").strip
 gem "rails", "6.1.3.2"
 
 gem "bootsnap"
+gem "dalli"
```
AlanGabbianelli, line 11: What is this used for? It'd be good to have an explanation in the commit message.

spec/support/helpers/oidc_client_helper.rb
AlanGabbianelli: I see.

> We'd need a double which only overrides one method and delegates all the others to a real instance

I wonder if that's a thing 🤔

AlanGabbianelli (approved):

PR #80, opened 2021-05-25
Fix pact branch & stub content store item

barrucadu:
Broken by #74

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -1,5 +1,8 @@
 require "webmock"
 require "pact/provider/rspec"
+require "gds_api/test_helpers/content_store"
+
+include GdsApi::TestHelpers::ContentStore
 
 def oidc_user
   OidcUser.find_or_create_by(sub: "user-id")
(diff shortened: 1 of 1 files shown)


spec/service_consumers/pact_helper.rb
```diff
@@ -79,6 +82,13 @@ def url_encode(str)
     allow(UserAttributes).to receive(:load_config_file).and_return(fixture_file)
 
     stub_request(:post, Plek.find("account-manager") + "/api/v1/jwt").to_return(status: 200, body: { id: "jwt-id" }.to_json)
+
+    %w[/guidance/some-govuk-guidance].each do |page_path|
+      stub_content_store_has_item(
+        page_path,
+        content_item_for_base_path(page_path).merge("content_id" => SecureRandom.uuid),
+      )
+    end
```
AlanGabbianelli, line 91: Why use an array here if we only have one item? Is this the same as
```ruby
page_path = "/guidance/some-govuk-guidance"

stub_content_store_has_item(
  page_path,
  content_item_for_base_path(page_path).merge("content_id" => SecureRandom.uuid),
)
```
?

barrucadu: Yeah, the array isn't needed.  I'll change it.

AlanGabbianelli (approved):

PR #85, opened 2021-05-26
Cache transition_checker_state locally

barrucadu:
At some point, we'll move the "Your Account" page from the
account-manager and onto GOV.UK proper (www.gov.uk/account ?).  When
that happens, this account-api will the only user of the
transition_checker_state attribute.

So to improve performance, we'd like to migrate the attribute from the
attribute-service into this app's database.  But we don't want to have
any downtime, so we can't go for the simple approach of:

1. Block updates to the attribute
2. Dump the attribute-service database
3. Import the attribute-service database into the account-api
4. Unblock updates to the attribute

As while the dump&import happens users won't be able to make changes.
Instead we want an online approach, which this commit enables:

1. Get account-api persisting transition_checker_state both
locally *and* remotely.

This is safe because account-api is the only *writer* of this
attribute value.  We don't need to keep anything else in sync.

Then later, we can do this:

2. Add an endpoint to attribute-service to expose the attribute values
3. Run a background task in account-api to fetch values which haven't
already been copied across by (1)
4. Stop account-api persisting the values remotely

This commit by itself is only the first step of the migration.  For
now, it will improve performance for users returning to their saved
results because, as values get cached locally, we won't have to query
the attribute-service to retrieve someone's saved results.

---

[Trello caerd](https://trello.com/c/gLfDdRtf/807-cache-transition-checker-answers-in-account-api)

### DIFF
--- a/app/lib/account_session.rb
+++ b/app/lib/account_session.rb
@@ -76,14 +76,30 @@ class AccountSession
 
   def get_attributes(attribute_names)
     local = attribute_names.select { |name| user_attributes.stored_locally? name }
-    remote = attribute_names.reject { |name| user_attributes.stored_locally? name }
+    remote = attribute_names.select { |name| !user_attributes.stored_locally?(name) && !user_attributes.cached_locally?(name) }
+    cached = attribute_names.select { |name| user_attributes.cached_locally? name }
 
-    get_local_attributes(local).merge(get_remote_attributes(remote)).compact
+    if cached
+      values_already_cached = get_local_attributes(cached)
+      values_to_cache = get_remote_attributes(cached.reject { |name| values_already_cached.key? name })
+      set_local_attributes(values_to_cache)
+      values = values_already_cached.merge(values_to_cache)
+    else
+      values = {}
+    end
+
+    values.merge(get_local_attributes(local).merge(get_remote_attributes(remote))).compact
   end
 
   def set_attributes(attributes)
     local = attributes.select { |name| user_attributes.stored_locally? name }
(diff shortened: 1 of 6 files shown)


app/lib/account_session.rb
```diff
@@ -76,14 +76,30 @@ def to_hash
 
   def get_attributes(attribute_names)
     local = attribute_names.select { |name| user_attributes.stored_locally? name }
-    remote = attribute_names.reject { |name| user_attributes.stored_locally? name }
+    remote = attribute_names.reject { |name| user_attributes.stored_locally? name }.reject { |name| user_attributes.cached_locally? name }
```
huwd, line 79: is rejecting these in sequence important for the result?

Otherwise could do?

`.reject { |name| user_attributes.stored_locally? name && user_attributes.cached_locally? name }`

no?

app/lib/account_session.rb
```diff
@@ -76,14 +76,30 @@ def to_hash
...
+    cached = attribute_names.reject { |name| user_attributes.stored_locally? name }.select { |name| user_attributes.cached_locally? name }
```
huwd, line 80: ditto here?

`.select { |name| user_attributes.stored_locally? name && user_attributes.cached_locally? name }`

huwd:
Otherwise looks good

app/lib/account_session.rb
barrucadu: Ok, I'll merge these.

huwd (approved):

PR #91, opened 2021-06-02
Refactor config code a little, add a notion of unwritable attributes, and make `email` and `email_verified` available

barrucadu:
This is so we can show the user's email address & prompt them to verify it on the account home page, which is going to move from the account manager to GOV.UK.

See commit messages for details.

---

[Trello card](https://trello.com/c/lIDQNATi/813-implement-the-govuk-account-home-page)

### DIFF
--- a/app/lib/account_session.rb
+++ b/app/lib/account_session.rb
@@ -75,9 +75,9 @@ class AccountSession
   end
 
   def get_attributes(attribute_names)
-    local = attribute_names.select { |name| user_attributes.stored_locally? name }
-    remote = attribute_names.select { |name| !user_attributes.stored_locally?(name) && !user_attributes.cached_locally?(name) }
-    cached = attribute_names.select { |name| user_attributes.cached_locally? name }
+    local = attribute_names.select { |name| user_attributes.type(name) == "local" }
+    remote = attribute_names.select { |name| user_attributes.type(name) == "remote" }
+    cached = attribute_names.select { |name| user_attributes.type(name) == "cached" }
 
     if cached
       values_already_cached = get_local_attributes(cached)
@@ -92,9 +92,9 @@ class AccountSession
   end
 
   def set_attributes(attributes)
-    local = attributes.select { |name| user_attributes.stored_locally? name }
-    remote = attributes.select { |name| !user_attributes.stored_locally?(name) && !user_attributes.cached_locally?(name) }
-    cached = attributes.select { |name| user_attributes.cached_locally? name }
+    local = attributes.select { |name| user_attributes.type(name) == "local" }
+    remote = attributes.select { |name| user_attributes.type(name) == "remote" }
+    cached = attributes.select { |name| user_attributes.type(name) == "cached" }
 
     if cached
       set_local_attributes(cached)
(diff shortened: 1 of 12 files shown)


app/lib/account_session.rb
```diff
@@ -75,9 +75,9 @@ def to_hash
   end
 
   def get_attributes(attribute_names)
-    local = attribute_names.select { |name| user_attributes.stored_locally? name }
-    remote = attribute_names.select { |name| !user_attributes.stored_locally?(name) && !user_attributes.cached_locally?(name) }
-    cached = attribute_names.select { |name| user_attributes.cached_locally? name }
+    local = attribute_names.select { |name| user_attributes.type(name) == "local" }
```
huwd, line 78: Interesting! I thought type was a reserved keywork in some cases?
I thought I'd had trouble with this before via ActiveRecord, but seems to be working here.

If we can get away with it type is a great and concise name!

app/lib/user_attributes.rb
```diff
@@ -1,5 +1,5 @@
 class UserAttributes
-  CONFIG_KEYS = %w[is_stored_locally is_cached_locally permissions].freeze
+  CONFIG_KEYS = %w[type permissions].freeze
```
huwd: Alternatively we could use an Enum here?

```
  enum :type, [ :local, :remote, :cached ]
```

That seems to let ActiveRecord give us a bunch of cool methods like `.remote?` or `.local?` or `.cached?`

app/lib/api_error/unwritable_attributes.rb
```diff
@@ -0,0 +1,25 @@
+module ApiError
+  class UnwritableAttributes < ApiError::Base
+    def initialize(attributes:)
+      super
+      @attributes = attributes
+    end
+
+    def status_code
+      :forbidden
```
huwd, line 9: makes sense!

app/lib/user_attributes.rb
```diff
@@ -1,11 +1,15 @@
 class UserAttributes
-  CONFIG_KEYS = %w[type permissions].freeze
+  CONFIG_KEYS = %w[type writable permissions].freeze
```
huwd, line 2: Again, I wonder if an enum would help us here?

Then we'd get `writeable?`

docs/api.md
```diff
@@ -408,6 +409,7 @@ Updates the attributes of the current user.
 #### Response codes
 
 - 422 if any attributes are unknown (see [error: unknown attribute names](#unknown-attribute-names))
+- 403 if any attributes are unwritable (see [error: unwritable attributes](#unwritable-attributes))
```
huwd, line 412: We'll need to update GDS API adaptors and possible frontend too right?

Makes me wonder a bit about piping all of this through frontend a bit... we've having to reproduce a lot of the error handling and replies there 🤔 especially when the api already uses problem details to send back a lot of good error data here.

barrucadu: No, the existing code won't need to change.  In practice, I don't think apps will need to worry about distinguishing the two types of 403 error, unless we start turning attributes which were once writable unwritable (but we shouldn't do that without letting people who expect to be able to update those attributes know).

huwd (approved):

PR #94, opened 2021-06-03
Add an API endpoint for getting batch user information

barrucadu:
This provides all the data needed for the new `/account/home` page, so
we can render that with a single request to account-api rather than
needing to make at least three separate requests to: get some
attributes, check other attributes, and check saved pages.

---

[Trello card](https://trello.com/c/lIDQNATi/813-implement-the-govuk-account-home-page)

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -1,0 +1,62 @@
+class UserController < ApplicationController
+  include AuthenticatedApiConcern
+
+  HOMEPAGE_ATTRIBUTES = %i[email email_verified transition_checker_state].freeze
+
+  def show
+    render_api_response(
+      {
+        level_of_authentication: @govuk_account_session.level_of_authentication,
+        email: attributes.dig(:email, :value),
+        email_verified: attributes.dig(:email_verified, :value),
+        services: {
+          transition_checker: attribute_service(:transition_checker_state),
+          saved_pages: saved_pages_service,
+        }.compact,
+      },
+    )
+  end
+
+private
+
+  def attribute_service(attribute_name)
+    attributes[attribute_name][:state]
+  end
+
+  def saved_pages_service
+    if @govuk_account_session.user.saved_pages.exists?
+      :yes
+    else
+      :no
+    end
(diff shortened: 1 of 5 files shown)


app/controllers/user_controller.rb
```diff
@@ -0,0 +1,62 @@
+class UserController < ApplicationController
+  include AuthenticatedApiConcern
+
+  HOMEPAGE_ATTRIBUTES = %w[email email_verified transition_checker_state].freeze
```
AlanGabbianelli, line 4: Could these be symbols (mainly for consistency in the `attributes.dig` arguments), or do we need them to be strings?

spec/requests/user_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    it "returns 'no'" do
+      get "/api/user", headers: headers
+      expect(service_state).to eq("no")
+    end
+
+    context "when the user has saved pages" do
+      before do
+        page_path = "/page-path"
+        stub_content_store_has_item(page_path, content_item_for_base_path(page_path).merge("content_id" => SecureRandom.uuid))
+        put saved_page_path(page_path: page_path), headers: headers
```
AlanGabbianelli, line 89: Can we put this in a `stub_user_has_saved_pages` method or similar?

spec/requests/user_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+  before do
+    stub_oidc_discovery
+
+    attributes.each do |name, value|
+      if value.nil?
+        stub_request(:get, "http://openid-provider/v1/attributes/#{name}").to_return(status: 404)
+      else
+        stub_request(:get, "http://openid-provider/v1/attributes/#{name}").to_return(body: { claim_value: value }.to_json)
+      end
+    end
```
AlanGabbianelli, line 15: Same as the previous comment: can we put this in a `stub_requests_for_attributes(attributes)` method or similar?
Mainly to abstract some of the complexity away from here.

AlanGabbianelli (approved):

PR #95, opened 2021-06-03
Enforce levels of authentication & remove support for legacy base64 sessions

barrucadu:
We've not yet enabled the ability to log in or register at level0 in
the account manager, so this won't impose any visible changes for
users who have logged in since late April.  Users who logged in before
then and still have a live session will be prompted to reauthenticate
when they next use the Transition Checker.

---

[Trello card](https://trello.com/c/6C8UdVFG/815-enforce-levels-of-authentication-in-production)

### DIFF
--- a/spec/requests/transition_checker_email_subscription_spec.rb
+++ b/spec/requests/transition_checker_email_subscription_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe "Transition Checker email subscriptions" do
   end
 
   let(:headers) { { "Content-Type" => "application/json", "GOVUK-Account-Session" => session_identifier } }
-  let(:session_identifier) { placeholder_govuk_account_session }
+  let(:session_identifier) { placeholder_govuk_account_session(level_of_authentication: "level1") }
 
   describe "GET" do
     before do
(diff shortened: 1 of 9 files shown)


spec/requests/transition_checker_email_subscription_spec.rb
```diff
@@ -66,27 +66,16 @@
     end
 
     context "when the user doesn't have a high enough level of authentication" do
-      let(:status) { 204 }
       let(:session_identifier) { placeholder_govuk_account_session(level_of_authentication: "level-1") }
 
-      it "checks if there is a subscription" do
+      it "returns a 403 and the required level" do
```
huwd, line 71: Do we need to updta the [docs](https://github.com/alphagov/account-api/blob/main/docs/api.md#get-apitransition-checker-email-subscription) too?

spec/requests/transition_checker_email_subscription_spec.rb
```diff
@@ -125,28 +114,14 @@
     context "when the user doesn't have a high enough level of authentication" do
       let(:session_identifier) { placeholder_govuk_account_session(level_of_authentication: "level-1") }
 
-      it "calls the account manager" do
-        stub = stub_request(:post, "#{Plek.find('account-manager')}/api/v1/transition-checker/email-subscription")
-          .with(body: hash_including(topic_slug: "slug"))
-          .to_return(status: 200)
-
+      it "returns a 403 and the required level" do
```
huwd, line 117: do we need to updat the [docs](https://github.com/alphagov/account-api/blob/main/docs/api.md#post-apitransition-checker-email-subscription) too?

huwd (approved):

spec/requests/transition_checker_email_subscription_spec.rb
barrucadu: This endpoint actually uses `level0` - so I'll remove this test (there's no such thing as `level-1`).  But the docs for the other endpoint do, yes

barrucadu: Actually... I guess it's not impossible we decide to bump the permission for checking up to `level1`.  So I'll leave the check in place and just update the docs.

PR #106, opened 2021-06-09
Add an endpoint to update a user's email by subject identifier

barrucadu:
I've called this `/api/oidc-users`, rather than using the existing
`/api/user`, because this endpoint feels pretty different to
`/api/user`:

- `/api/user` is for the user's profile details as used on GOV.UK
- the new endpoint is for the identity provider to update special
  details (currently email)

They're both about the user, but through different lenses: the user as
seen on GOV.UK, and the user as managed by the OIDC service.  Hence,
`/api/oidc-users`.

You'll need to re-seed your database after pulling this change.

---

[Trello card](https://trello.com/c/UD2IR18h/830-make-account-manager-tell-account-api-about-email-address-changes)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,8 @@ Rails.application.routes.draw do
 
     get "/user", to: "user#show"
 
+    resources :oidc_users, only: %i[update], param: :subject_identifier, path: "oidc-users"
+
     get "/attributes", to: "attributes#show"
     patch "/attributes", to: "attributes#update"
 
--- a/spec/requests/oidc_users_spec.rb
+++ b/spec/requests/oidc_users_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe "OIDC Users endpoint" do
 
   describe "PUT" do
     it "creates the user if they do not exist" do
-      expect { put oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.to change(OidcUser, :count)
+      expect { put oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.to change(OidcUser, :count).by(1)
       expect(response).to be_successful
     end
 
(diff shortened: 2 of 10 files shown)


config/routes.rb
```diff
@@ -14,6 +14,8 @@
 
     get "/user", to: "user#show"
 
+    resources :oidc_users, only: %i[update], param: :subject_identifier, path: "oidc-users"
```
AlanGabbianelli, line 17: Is the path specified here to have `/api/oidc-users` instead of `/api/oidc_users`?

barrucadu: Yes

AlanGabbianelli: What's the advantage of using `resources` if we only need `update`?

spec/requests/oidc_users_spec.rb
```diff
@@ -0,0 +1,42 @@
+RSpec.describe "OIDC Users endpoint" do
+  let(:headers) { { "Content-Type" => "application/json" } }
+  let(:params) { { email: email, email_verified: email_verified }.compact.to_json }
+  let(:email) { "email@example.com" }
+  let(:email_verified) { true }
+  let(:subject_identifier) { "subject-identifier" }
+
+  describe "PUT" do
+    it "creates the user if they do not exist" do
+      expect { put oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.to change(OidcUser, :count)
```
AlanGabbianelli, line 10:
```suggestion
      expect { put oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.to change(OidcUser, :count).by(1)
```

config/routes.rb
barrucadu: I can imagine us wanting a `delete` endpoint in the future too.  But even if we don't, using `resources` seems like a good way to keep in mind the principle that controllers should only have the CRUD actions where possible.

AlanGabbianelli (approved):

PR #111, opened 2021-06-11
Implement email subscriptions

barrucadu:
This PR adds an API for email subscription management, and makes it so that when the Identity Provider tells us that a user has a new email address, we update the subscriptions.

**Why not just use email-alert-api?**

Updating the subscriptions when the account updates needs a change in account-api, but couldn't we just use email-alert-api for the rest of this?  Why do we need an `email_subscriptions` table?  Well, there are two reasons:

1. Since consuming apps work with "names" rather than email topic slugs, they don't need to implement logic like:
    1. Call email-alert-api with the old topic slug and the current email address to cancel the old subscription (this would be a new endpoint in email-alert-api)
    2. Call email-alert-api with the new topic slug and current email address to create the new subscription
2. We're intentionally *not* fully integrating accounts with notifications yet, so when a user's email address is updates we *only* want to update those subscriptions associated with the account (we could do this by adding a flag to the email-alert-api model)

So we can keep all the changes in account-api, or we can make some in account-api and some in email-alert-api.  Probably best to keep them all in account-api for now.  We might revisit this after integrating accounts and notifications more tightly.

---

[Trello card](https://trello.com/c/yzVmdrzw/829-add-rest-api-for-email-subscriptions-to-account-api)

### DIFF
--- a/app/controllers/email_subscriptions_controller.rb
+++ b/app/controllers/email_subscriptions_controller.rb
@@ -1,0 +1,60 @@
+class EmailSubscriptionsController < ApplicationController
+  include AuthenticatedApiConcern
+
+  before_action :check_subscription_exists, only: %i[show destroy]
+
+  def show
+    if email_subscription.email_alert_api_subscription_id
+      begin
+        state = GdsApi.email_alert_api.get_subscription(email_subscription.email_alert_api_subscription_id)
+        if state.to_hash.dig("subscription", "ended_reason")
+          email_subscription.destroy!
+          head :not_found and return
+        end
+      rescue GdsApi::HTTPGone, GdsApi::HTTPNotFound
+        email_subscription.destroy!
+        head :not_found and return
+      end
+    end
+
+    render_api_response(email_subscription: email_subscription.to_hash)
+  end
+
+  def update
+    attributes = @govuk_account_session.get_attributes(%w[email email_verified])
+
+    email_subscription = EmailSubscription.transaction do
+      EmailSubscription
+        .create_with(topic_slug: params.fetch(:topic_slug))
+        .find_or_create_by!(
+          oidc_user_id: @govuk_account_session.user.id,
+          name: params.fetch(:subscription_name),
+        ).tap { |subscription| subscription.update!(topic_slug: params.fetch(:topic_slug)) }
+    end
+
+    email_subscription.reactivate_if_confirmed!(
+      attributes["email"],
+      attributes["email_verified"],
+    )
+
+    render_api_response(email_subscription: email_subscription.to_hash)
+  end
+
+  def destroy
+    email_subscription.destroy!
+    head :no_content
+  end
+
+private
+
+  def email_subscription
+    @email_subscription ||= EmailSubscription.find_by(
+      oidc_user: @govuk_account_session.user,
+      name: params.fetch(:subscription_name),
+    )
+  end
+
+  def check_subscription_exists
+    head :not_found and return unless email_subscription
+  end
+end
--- a/app/models/email_subscription.rb
+++ b/app/models/email_subscription.rb
@@ -1,0 +1,14 @@
+class EmailSubscription < ApplicationRecord
+  belongs_to :oidc_user
+
+  validates :name, presence: true
+  validates :topic_slug, presence: true
+
+  def to_hash
+    {
+      "name" => name,
+      "topic_slug" => topic_slug,
+      "email_alert_api_subscription_id" => email_alert_api_subscription_id,
+    }.compact
+  end
+end
--- a/app/models/email_subscription.rb
+++ b/app/models/email_subscription.rb
@@ -4,11 +4,43 @@ class EmailSubscription < ApplicationRecord
   validates :name, presence: true
   validates :topic_slug, presence: true
 
+  before_destroy :deactivate!
+
   def to_hash
     {
       "name" => name,
       "topic_slug" => topic_slug,
       "email_alert_api_subscription_id" => email_alert_api_subscription_id,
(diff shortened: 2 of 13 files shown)


app/models/email_subscription.rb
```diff
@@ -4,11 +4,42 @@ class EmailSubscription < ApplicationRecord
...
+
+    update!(email_alert_api_subscription_id: subscription.to_hash.dig("subscription", "id"))
+  end
+
+  def deactivate!
+    return unless email_alert_api_subscription_id
+
+    GdsApi.email_alert_api.unsubscribe(email_alert_api_subscription_id)
+    update!(email_alert_api_subscription_id: nil)
+  rescue GdsApi::HTTPGone, GdsApi::HTTPNotFound
```
huwd, line 41: So in this flow,
- We want to deactivate a user before destroying an account
- If the call to `GdsApi.email_alert_api.unsubscribe` returns either `gone` or `not found` then this is rescued and the destroy continues anyway?

Are there any other status codes we would wish to check for?
Is it any better if we wrapped like 40 in something like?

```
if response.code == 200
   update!(email_alert_api_subscription_id: nil)
end
```

Not sure what other possible return values from the API would be, and when it's OK to delete a subscription if we've not sucessfulyl unsubscribed a user?

app/models/email_subscription.rb
```diff
@@ -4,11 +4,42 @@ class EmailSubscription < ApplicationRecord
...
+    return unless email_verified
+
+    subscriber_list = GdsApi.email_alert_api.get_subscriber_list(
+      slug: topic_slug,
+    )
+
+    subscription = GdsApi.email_alert_api.subscribe(
+      subscriber_list_id: subscriber_list.to_hash.dig("subscriber_list", "id"),
+      address: email,
+      frequency: "daily",
```
huwd, line 29: Are we happy with "daily" being the default here?
Gavin & co might be worth checking in with, I think there was some work on notifications previously to make a weekly digest the default as there were complaints from users about getting too many emails from default settings?

Not sure if that's still a valid concern here?

spec/models/email_subscription_spec.rb
```diff
@@ -9,4 +13,83 @@
...
+  describe "reactivate_if_confirmed!" do
+    let(:email) { "email@example.com" }
+    let(:email_verified) { false }
+
+    it "doesn't call email-alert-api if the user is not confirmed" do
+      stub = stub_subscriber_list
+
+      email_subscription.reactivate_if_confirmed!(email, email_verified)
+
+      expect(stub).not_to have_been_made
```
huwd, line 26: That's a cool expectation, not used that before. Will do so more often!

app/controllers/email_subscriptions_controller.rb
```diff
@@ -0,0 +1,60 @@
+class EmailSubscriptionsController < ApplicationController
+  include AuthenticatedApiConcern
+
+  before_action :check_subscription_exists, only: %i[show destroy]
+
+  def show
+    if email_subscription.email_alert_api_subscription_id
```
huwd, line 7: Do we want any kind of error response if we can't find a subscription by the `subscription_name`?
I guess if the header is invalid they'll be logged out.

Is the problem detail response of any use here?

spec/service_consumers/pact_helper.rb
```diff
@@ -1,9 +1,35 @@
 require "webmock"
 require "pact/provider/rspec"
+require "plek"
```
huwd, line 3: Interesting, I don't see anywhere new `Plek` is being used here.
There are the existing lines on [L123](https://github.com/alphagov/account-api/pull/111/commits/d36ee3d2608ff1c426a15ecdc07c4ec1f4009c9a#diff-7d17ffbce13a8dc8820ff122a86f9a63284e76991d64a9981a7eb7c5697f6bfcR123) and [L124](https://github.com/alphagov/account-api/pull/111/commits/d36ee3d2608ff1c426a15ecdc07c4ec1f4009c9a#diff-7d17ffbce13a8dc8820ff122a86f9a63284e76991d64a9981a7eb7c5697f6bfcR124)


Were those broken previously?

app/models/email_subscription.rb
barrucadu: Daily is what we currently use for the transition checker subscriptions, so I just copied that over

app/controllers/email_subscriptions_controller.rb
barrucadu: There's the `before_action` which will return a 404

spec/service_consumers/pact_helper.rb
barrucadu: Without this. an error gets thrown by the `require "gds_api/test_helpers/email_alert_api"` line:

```
uninitialized constant GdsApi::TestHelpers::EmailAlertApi::Plek
```

So I think:

- Below, where we use `Plek`, the Rails constant autoloading was finding and pulling in the module
- Here, the reference to `Plek` inside the test helper is *not* resolved to `::Plek`, and so it throws an error

Bit weird.

app/models/email_subscription.rb
barrucadu: It returns a 204 if the subscription has been ended, and a 404 or 410 if subscription has been ended through some other mechanism (like the user cancelling it themselves through email-alert-frontend).

So actually this code is slightly wrong, it should consider a 404 and a 410 success too - I'll fix the `update!` and add a test.

barrucadu: f80df855ae3517220fe616545d166789795e1a6f

huwd (approved):

PR #113, opened 2021-06-16
Reorganise OIDC spec helpers a bit

barrucadu:
Our API coverage looks pretty good, it's really only the `redirect_path` of the auth controller and the `page_path` of the saved pages controller where we're passing something which could directly come from user input.  And those have good coverage now.

I shuffled around the OIDC helpers a bit, merging two single-use methods into the file which needed them and adding a couple of new methods for stubbing remote attribute requests.

---

[Trello card](https://trello.com/c/siLdMmuu/848-review-api-tests-for-missing-coverage)

### DIFF
--- a/spec/support/helpers/oidc_client_helper.rb
+++ b/spec/support/helpers/oidc_client_helper.rb
@@ -30,6 +30,18 @@ module OidcClientHelper
     allow_any_instance_of(OidcClient).to receive(:tokens!).and_return(token_response)
     # rubocop:enable RSpec/AnyInstance
   end
+
+  def stub_remote_attribute_request(name:, value: nil, status: nil)
+    status ||= value.nil? ? 404 : 200
+    stub_request(:get, "http://openid-provider/v1/attributes/#{name}")
+      .to_return(status: status, body: { claim_value: value }.compact.to_json)
+  end
+
+  def stub_remote_attribute_requests(names_and_values = {})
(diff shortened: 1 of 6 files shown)


spec/support/helpers/oidc_client_helper.rb
```diff
@@ -30,6 +30,18 @@ def stub_token_response
...
+
+  def stub_remote_attribute_request(name:, value: nil, status: nil)
+    status ||= value.nil? ? 404 : 200
+    stub_request(:get, "http://openid-provider/v1/attributes/#{name}")
+      .to_return(status: status, body: { claim_value: value }.compact.to_json)
+  end
+
+  def stub_remote_attribute_requests(names_and_values = {})
+    names_and_values.each do |name, value|
+      stub_remote_attribute_request(name: name, value: value)
```
huwd, line 42: So we're happy that these can only ever 200?
I guess this is the most high-use stub for a happy path?

Thinking out loud:
If we wanted to test 422 or 404 across multiple (... would we?) we can write a specific stub for those...

huwd (approved):

spec/support/helpers/oidc_client_helper.rb
barrucadu: Yeah, we could always make a new stub for a situation like that, or extend this one

PR #117, opened 2021-06-17
Add `has_unconfirmed_email` attribute, to be used in the account home page

barrucadu:
This is not just `!email_verified`:

- `email` - this is the email address the user registered with, or a
  new email address they have confirmed.

- `email_verified` - the user needs to verify the email address they
  registered with, this starts `false`, becomes `true`, and then never
  returns to `false`

- `has_unconfirmed_email` - this is `true` if `email_verified` is
  `false` OR the user has a pending email change to confirm

So,

    if has_unconfirmed_email
      if email_verified
        show "finish changing your email address" banner
      else
        show "confirm your account" banner
      end
    end

And set the attribute through `PATCH /api/oidc-users/:sub`.

We want to have the as many of the attributes needed to return the
`GET /api/user` response stored locally as possible, so the homepage
is nice and fast.

---

[Trello card](https://trello.com/c/nvi8bZct/856-add-hasunconfirmedemail-attribute)

### DIFF
--- a/spec/requests/user_spec.rb
+++ b/spec/requests/user_spec.rb
@@ -16,6 +16,7 @@ RSpec.describe "User information endpoint" do
     {
       email: "email@example.com",
       email_verified: true,
+      has_unconfirmed_email: false,
       transition_checker_state: transition_checker_state,
     }
   end
--- a/app/controllers/oidc_users_controller.rb
+++ b/app/controllers/oidc_users_controller.rb
@@ -1,5 +1,5 @@
 class OidcUsersController < ApplicationController
-  OIDC_USER_ATTRIBUTES = %i[email email_verified].freeze
+  OIDC_USER_ATTRIBUTES = %i[email email_verified has_unconfirmed_email].freeze
 
   def update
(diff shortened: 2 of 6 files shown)


spec/requests/user_spec.rb
```diff
@@ -16,6 +16,7 @@
     {
       email: "email@example.com",
       email_verified: true,
```
huwd, line 18: given this is a one-way-switch...
is there any utility in making this someting more like:
`email_verified_on: "2021-01-01"`, where `nil` means never verified?
Wonder if that is a better form of history than `true` or `false?

Whereas `has_unconfirmed_email` is a representation of present state?

huwd: atch, but then we'd have to update `email_verified_at` and perhaps that's just another thing that could go wrong. Probably nevermind!

huwd (approved):

spec/requests/user_spec.rb
barrucadu: Another option might be to have a three state thing:

```
email_state = :confirmed
email_state = :pending_reconfirmation
email_state = :pending_confirmation
```

PR #127, opened 2021-06-29
Update linked subscribers on email address change

barrucadu:
Currently, if a user updates their email address, any subscriptions
which account-api knows about will also be updated.  But any other
subscriptions for the same email address won't.  This isn't great.

To address this problem, we've added a concept of "linking" a
subscriber to a GOV.UK Account in email-alert-api.

So now the logic for updating email subscriptions is:

- Check if there is a linked subscriber:
  - If so, update the address for that subscriber
  - If not, update each subscription we know about

We're going to make some frontend changes to nudge people towards
linking their accounts.

---

[Trello card](https://trello.com/c/Vzd8so2T/874-implement-designs-for-managing-your-email-notifications-via-the-account)

### DIFF
--- a/app/controllers/oidc_users_controller.rb
+++ b/app/controllers/oidc_users_controller.rb
@@ -6,12 +6,35 @@ class OidcUsersController < ApplicationController
     user.set_local_attributes(params.permit(OIDC_USER_ATTRIBUTES).to_h)
     attributes = user.get_local_attributes(OIDC_USER_ATTRIBUTES)
 
-    if attributes["email"] && !attributes["email_verified"].nil?
-      user.email_subscriptions.each do |email_subscription|
-        email_subscription.reactivate_if_confirmed!(
-          attributes["email"],
-          attributes["email_verified"],
+    if attributes["email"] && attributes["email_verified"]
+      begin
+        # if the user has linked their notifications account to their
+        # GOV.UK account we don't need to update their
+        # `user.email_subscriptions`, because we can update the
+        # subscriber directly.
+        subscriber_id = GdsApi.email_alert_api
+          .find_subscriber_by_govuk_account(govuk_account_id: user.id)
+          .dig("subscriber", "id")
+        GdsApi.email_alert_api.change_subscriber(
+          id: subscriber_id,
+          new_address: attributes["email"],
         )
+      rescue GdsApi::HTTPNotFound
+        # but for users who haven't linked their notifications account
+        # to their GOV.UK account, we do need to update any
+        # account-linked subscriptions they have (eg, brexit checker
+        # users who haven't touched notifications since registering
+        # through the checker).
+        #
+        # this branch can be removed once we have no GOV.UK accounts
+        # which have subscriptions but are *not* linked to the
+        # corresponding notifications account.
+        user.email_subscriptions.each do |email_subscription|
+          email_subscription.reactivate_if_confirmed!(
+            attributes["email"],
+            attributes["email_verified"],
+          )
+        end
       end
     end
 
--- a/spec/requests/oidc_users_spec.rb
+++ b/spec/requests/oidc_users_spec.rb
@@ -11,6 +11,11 @@ RSpec.describe "OIDC Users endpoint" do
   let(:subject_identifier) { "subject-identifier" }
 
   describe "PUT" do
+    before do
+      next_oidc_user_id = ActiveRecord::Base.connection.execute("select last_value from oidc_users_id_seq").first["last_value"] + 1
+      stub_email_alert_api_find_subscriber_by_govuk_account_no_subscriber(next_oidc_user_id)
+    end
+
     it "creates the user if they do not exist" do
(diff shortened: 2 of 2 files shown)


app/controllers/oidc_users_controller.rb
```diff
@@ -6,12 +6,35 @@ def update
...
+        # subscriber directly.
+        subscriber_id = GdsApi.email_alert_api
+          .find_subscriber_by_govuk_account(govuk_account_id: user.id)
+          .dig("subscriber", "id")
+        GdsApi.email_alert_api.change_subscriber(
+          id: subscriber_id,
+          new_address: attributes["email"],
         )
+      rescue GdsApi::HTTPNotFound
+        # but for users who haven't linked their notifications account
```
huwd, line 23: This seems like quite the oversight.
Are we sure that wasn't deliberate?

I can't see why it would be though 🤔 ...
This comes straight after a notification subscriber has proven access to an email account right?
So there aren't any authentication benefits... we'd just be asking the user to repeat the process N times...

And if folks had a new email address or wanted to only update some... they could just create a new (perhaps a "work") account.

So this looks good to me!

huwd: I wonder how many subscriptions the average user has now and how frequently they update?
Any chance we'll need more workers if this becomes a regular operation?

spec/requests/oidc_users_spec.rb
```diff
@@ -11,6 +11,11 @@
   let(:subject_identifier) { "subject-identifier" }
 
   describe "PUT" do
+    before do
+      next_oidc_user_id = ActiveRecord::Base.connection.execute("select last_value from oidc_users_id_seq").first["last_value"] + 1
```
huwd, line 15: This seems like an odd query, can you unpack this line for me and why we need to do it this way?
I read that as:
```
- use an active record method to:
- grab the sequentially highest oidc user id?
- access the last value
- add one to it...
```

Why are we doing this instead of using  a factory / fixture or a mock?

huwd, line 15: This seems like an odd query, can you unpack this line for me and why we need to do it this way?
I read that as:
```
- use an active record method to:
- grab the sequentially highest oidc user id?
- access the last value
- add one to it...
```

Why are we doing this instead of using  a factory / fixture or a mock?

app/controllers/oidc_users_controller.rb
barrucadu: I'm not sure what you mean, what oversight?  This endpoint is called when a user changes the email address of their GOV.UK Account, not their Notifications Account.

There are 3 possible states a user could be in:

1. A user has a GOV.UK Account and a Notifications Account and they are linked
2. A user has a GOV.UK Account and a Notifications Account and they are not linked
3. A user has a GOV.UK Account and no Notifications Account

So, we call email-alert-api to see if there is a Notifications Account linked to this GOV.UK Account.  If there is, we're in case 1 and can update the Notifications Account directly.

If there isn't, we're in either case 2 or 3.  And in both cases the behaviour is the same: update the address of only the email subscriptions which account-api knows about (eg, the Transition Checker subscription), because the expected behaviour of those subscriptions is that they are linked to the account.

account-api doesn't have a database of all subscriptions, only those which are explicitly connected to the account.  Which is currently just the Transition Checker subscription, but in the future may also include others.

huwd: > But any other subscriptions for the same email address won't. This isn't great.

ah ok sorry I thought this meant that the current behaviour of the notifications was that an unlinked Notifications account can only update the email address on a per-subscription list basis?

So the commit message isn't making that comparison then?

spec/requests/oidc_users_spec.rb
huwd: ah i see so we're getting the range and then going one beyond it to ensure there's never any clash?

huwd (approved):

app/controllers/oidc_users_controller.rb
barrucadu: Individual subscriptions don't have an email address associated with them, they belong to a Notifications Account, and a Notifications Account has an email address.

So the current behaviour for unlinked accounts where we do nevertheless have an individual subscription which account-api knows about (like the brexit checker subscriptions), is to delete the old one and make a new one for the new email address - recreating the subscription under a *different* Notifications Account*.

But when a Notifications Account and a GOV.UK Account are linked, we don't need to move subscriptions between Notifications Accounts, we can just update the Notifications Account itself.  Does that help?

spec/requests/oidc_users_spec.rb
barrucadu: It's because the account-api user ID is passed as a parameter to email-alert-api, but we want to test the case where the account-api user doesn't exist yet.  Which means it'll create a new user than then call email-alert-api with that user's ID - so to stub the request, we need to know what that ID will be.

PR #136, opened 2021-07-07
Incrementally migrate Transition Checker subscriptions

barrucadu:
All Transition Checker email subscription interactions are now going
through this controller, and the JWT method of creating subscriptions
is gone.

So now we can begin to migrate the state into account-api.

---

[Trello card](https://trello.com/c/5rNWhML7/885-begin-incremental-migration-of-transition-checker-email-subscriptions-from-account-manager-to-account-api)

### DIFF
--- a/app/controllers/email_subscriptions_controller.rb
+++ b/app/controllers/email_subscriptions_controller.rb
@@ -3,36 +3,10 @@ class EmailSubscriptionsController < ApplicationController
 
   TRANSITION_CHECKER_SUBSCRIPTION_NAME = "transition-checker-results".freeze
 
+  before_action :migrate_transition_checker_subscription
   before_action :check_subscription_exists, only: %i[show destroy]
 
-  # the transition checker-specific stuff in here is temporary: after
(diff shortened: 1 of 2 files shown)


app/controllers/email_subscriptions_controller.rb
```diff
@@ -82,12 +44,30 @@ def update
...
+    email_subscription.destroy!
     head :no_content
   end
 
 private
 
+  # remove this after we've done a bulk import of older subscriptions
+  def migrate_transition_checker_subscription
+    return unless params.fetch(:subscription_name) == TRANSITION_CHECKER_SUBSCRIPTION_NAME
+    return if @email_subscription
```
barrucadu, line 56: This is wrong, should be `return if email_subscription` - I think a test for the case where the subscription has already been migrated would catch this.

barrucadu (commented):
I thought I'd had this one reviewed last week, but I must have missed it 😬

theseanything (approved):
Looks good! Nit: I'd update the commit message with more of a description on what is being migrated and why.

PR #140, opened 2021-07-12
Record timer metrics from OidcClient

barrucadu:
Some of the account-api endpoints have a pretty high variance in
response time.  Since the logic in account-api itself is not too
complex, my hunch is that this is the attribute service or account
manager being slow.  So, record some timer metrics to gather some more
evidence.

If it is the OIDC stuff being slow, then maybe we can add more
instances there.  And if it's not, well, this will still be useful
debugging data to have for future problems.

---

[Trello card](https://trello.com/c/UPXoxbRt/904-add-timing-metrics-to-account-api)

### DIFF
--- a/app/lib/oidc_client.rb
+++ b/app/lib/oidc_client.rb
@@ -31,7 +31,9 @@ class OidcClient
   def callback(auth_request, code)
     client.authorization_code = code
 
-    tokens!(oidc_nonce: auth_request.oidc_nonce)
+    time_and_return "tokens" do
+      tokens!(oidc_nonce: auth_request.oidc_nonce)
+    end
   end
 
   def tokens!(oidc_nonce: nil)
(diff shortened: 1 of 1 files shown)


app/lib/oidc_client.rb
```diff
@@ -53,12 +55,14 @@ def tokens!(oidc_nonce: nil)
...
 
   def userinfo(access_token:, refresh_token:)
-    response = oauth_request(
-      access_token: access_token,
-      refresh_token: refresh_token,
-      method: :get,
-      uri: userinfo_endpoint,
-    )
+    response = govuk_timer "userinfo" do
+      oauth_request(
```
theseanything, line 59: Nit: I found the nested block somewhat difficult to read as it's not obvious the govuk_timer just returns the result of the block. Not sure if it's possible to set the attribute within the block instead?

```suggestion
    govuk_timer "userinfo" do
      response = oauth_request(
```

(This shouldn't block this PR)

theseanything (approved):

app/lib/oidc_client.rb
barrucadu: How about if I changed the method name to `time_and_return`?

theseanything: Yup, that'd be improvement!

theseanything (approved):

PR #141, opened 2021-07-13
Fetch remote attributes from userinfo, if present there

barrucadu:
The userinfo response MUST have at least one attribute, "sub", the
user's "subject identifier", and MAY also have additional attributes.

In contrast, the `get_attribute` endpoint in our attribute-service
prototype only returns one attribute at a time.  So an account-api
request which needs to fetch multiple attributes (like `GET /api/user`)
will make multiple requests, in sequence, to fetch them.
Which is slow.

If we can get even one attribute from the userinfo response (and in
practice we can currently get *all* of them from the userinfo
response), then this change is worthwhile.

---

[Trello card](https://trello.com/c/RYLdsABd/906-fetch-remote-attributes-from-userinfo-if-present-there)

### DIFF
--- a/app/lib/account_session.rb
+++ b/app/lib/account_session.rb
@@ -14,7 +14,7 @@ class AccountSession
     @level_of_authentication = level_of_authentication
     @frozen = false
 
-    @user_id = user_id || oidc_do(:userinfo)["sub"]
+    @user_id = user_id || userinfo["sub"]
   end
 
   def self.deserialise(encoded_session:, session_signing_key:)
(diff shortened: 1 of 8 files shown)


app/lib/account_session.rb
```diff
@@ -103,7 +103,15 @@ def migrate_transition_checker_email_subscription
   attr_reader :session_signing_key
 
   def get_remote_attributes(remote_attributes)
-    remote_attributes.index_with { |name| oidc_do :get_attribute, { attribute: name } }.compact
+    values = remote_attributes.index_with do |name|
+      if userinfo.key? name
+        userinfo[name]
+      else
+        oidc_do(:get_attribute, { attribute: name })
```
bevanloon, line 110: nitpick - feel free to ignore. Looks like we usually leave off the brackets in the rest of this class when calling `oidc_do`

bevanloon (approved):

PR #147, opened 2021-07-14
Move the deletion of expired AuthRequests to a sidekiq worker

barrucadu:
The app is doing a lot of garbage collection, and specifically it
looks like it's allocating a lot of ActiveRecord transaction objects.

I suspect that that's caused by this in the `sign_in` controller:

    AuthRequest.expired.delete_all

Deleting expired objects on every request was implemented as an easy
"is this really worth setting up sidekiq for?" approach but, if it's
causing excessive allocation, and so excessive garbage collection,
it *is* worth setting up sidekiq to handle.

ActiveJob was already present in the app but I've removed it
completely (rather than configure it to use sidekiq as a queuing
system) because on GOV.UK we're more experienced with using sidekiq
directly.

See also https://github.com/alphagov/govuk-puppet/pull/11200

---

[Trello card](https://trello.com/c/uRW8OS61/907-move-expired-authrequest-deletion-to-sidekiq)

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,3 +3,8 @@
 
 :queues:
   - default
+
+:schedule:
+  expired_auth_request_worker:
+    every: '1h'
+    class: ExpiredAuthRequestWorker
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,0 +1,5 @@
+---
+:logfile: ./log/sidekiq.json.log
+
+:queues:
+  - default
--- a/Gemfile
+++ b/Gemfile
@@ -13,9 +13,11 @@ gem "gds-api-adapters"
 gem "gds-sso"
 gem "govuk_app_config"
 gem "govuk_message_queue_consumer"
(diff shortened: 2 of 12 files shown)


config/sidekiq.yml
```diff
@@ -0,0 +1,3 @@
+---
+:queues:
+  - default
```
bevanloon, line 3: I notice in a few of our other repos, we set a `:logfile: ./log/sidekiq.json.log`. Is that something we want here? I'm guessing the implications of not having it means we don't explicitly log to file (based on https://github.com/alphagov/govuk_sidekiq/blob/master/lib/govuk_sidekiq/sidekiq_initializer.rb#L32), but I don't know our sidekiq setup very well tbh.

barrucadu: I'm not sure what sidekiq logs by default, given that this new worker isn't making any logging calls, but I guess adding a `:logfile:` can't hurt

bevanloon (approved):

PR #154, opened 2021-07-20
Delete users

huwd:
A part of: [Trello](https://trello.com/c/b95YV8qC/912-%E2%9D%97-delete-users-in-account-api-when-theyre-deleted-in-account-manager)

We are currently migrating features from our PaaS based
[account-manager-prototype](https://github.com/alphagov/govuk-account-manager-prototype)
to account-api.

Accounts API now acts as an additional store of user data. When a user
deletes their account through the account manager we require an
authenticated endpoint that will forward the request to account-api so
we can remove their data there too.

### DIFF
--- a/app/controllers/oidc_users_controller.rb
+++ b/app/controllers/oidc_users_controller.rb
@@ -42,6 +42,14 @@ class OidcUsersController < ApplicationController
     render json: attributes.merge(sub: user.sub)
   end
 
+  def destroy
+    user = OidcUser.find_by!(sub: params.fetch(:subject_identifier))
+    user.destroy!
+    head :no_content and return
+  rescue ActiveRecord::RecordNotFound
+    head :not_found and return
+  end
+
 private
 
   def authorise_sso_user!
--- a/docs/api.md
+++ b/docs/api.md
@@ -12,6 +12,7 @@ management. This API is not for other government services.
   - [`GET /api/oauth2/sign-in`](#get-apioauth2sign-in)
   - [`POST /api/oauth2/callback`](#post-apioauth2callback)
   - [`GET /api/user`](#get-apiuser)
+  - [`DELETE /api/user`](#delete-apiuser)
   - [`GET /api/attributes`](#get-apiattributes)
   - [`PATCH /api/attributes`](#patch-apiattributes)
   - [`GET /api/attributes/names`](#get-apiattributesnames)
@@ -224,6 +225,33 @@ Response:
     }
 }
 ```
+
+### `DELETE /api/user`
+
+Removes a user's account.
+
+#### Request headers
+
+- `GOVUK-Account-Session`
+  - the user's session identifier
+
+#### Response codes
+
+- 401 if the session identifier is invalid
+- 404 if the user cannot be found
+- 204 otherwise
+
+#### Example request / response
+
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.account_api.delete_user(
+    govuk_account_session: "session-identifier",
+)
+```
+
+Response is status code only.
 
 ### `GET /api/attributes`
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,7 +14,7 @@ Rails.application.routes.draw do
 
     get "/user", to: "user#show"
 
-    resources :oidc_users, only: %i[update], param: :subject_identifier, path: "oidc-users"
+    resources :oidc_users, only: %i[update destroy], param: :subject_identifier, path: "oidc-users"
 
     get "/attributes", to: "attributes#show"
     patch "/attributes", to: "attributes#update"
--- a/spec/requests/oidc_users_spec.rb
+++ b/spec/requests/oidc_users_spec.rb
@@ -100,7 +100,9 @@ RSpec.describe "OIDC Users endpoint" do
 
     context "when a user does not exist" do
       it "does not change the count of users and returns not found" do
-        expect { delete oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.not_to change(OidcUser, :count)
+        without_detailed_exceptions do
+          expect { delete oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.not_to change(OidcUser, :count)
+        end
         expect(response).to be_not_found
       end
     end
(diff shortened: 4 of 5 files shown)


app/controllers/oidc_users_controller.rb
```diff
@@ -42,6 +42,14 @@ def update
     render json: attributes.merge(sub: user.sub)
   end
 
+  def destroy
+    user = OidcUser.find_by!(sub: params.fetch(:subject_identifier))
+    user.destroy!
+    head :no_content and return
+  rescue ActiveRecord::RecordNotFound
+    head :not_found and return
```
huwd, line 50: I'm unsure if this is needed or not.
On the one hand we're doing a search here, and if we don't rescue then we'd return a `500` instead of a `404` which seems less descriptive to me.

But on the other, I'm not sure if tis is a code path that should ever be reachable.
A user should not be able to submit a subject identifier other than their own... which should be present...

If folks think this is OTT, can remove it.

aldavidson (changes requested):
Looks fine, but see my reply about the `rescue ... :not_found` clause

app/controllers/oidc_users_controller.rb
aldavidson: > A user should not be able to submit a subject identifier other than their own..
True, _but_ that doesn't guarantee that it's present at the time the request is processed. There may be multiple requests in parallel, or async job retries, or someone doing an operation in the console just at the time the users request makes it through the queue, and so on.

That being said, Rails should rescue any `ActiveRecord::NotFound` error in a controller with a 404 by default (see the note about `config.action_dispatch.rescue_responses` in [Configuring Rails Apps](https://guides.rubyonrails.org/configuring.html#configuring-action-dispatch). It looks like we haven't done anything to alter this default -  from the console, run:
```
> Rails.configuration.action_dispatch.rescue_responses
=> {"ActiveRecord::RecordNotFound"=>:not_found,
 "ActiveRecord::StaleObjectError"=>:conflict,
 "ActiveRecord::RecordInvalid"=>:unprocessable_entity,
 "ActiveRecord::RecordNotSaved"=>:unprocessable_entity}
```

so I think we should be ok without this rescue clause. You could always add a test case for this?

huwd: Right ok yeah so too broad in the rescue, that makes sense. I'll simplify this!

huwd: 🤔 humm so when I remove this i get:
```
Failures:                                                                                                                                                                                                          
                                                                                                                                                                                                                   
  1) OIDC Users endpoint DELETE when a user does not exist does not change the count of users and returns not found                                                                                                
     Failure/Error: user = OidcUser.find_by!(sub: params.fetch(:subject_identifier))                                                                                                                               
                                                                                                                                                                                                                   
     ActiveRecord::RecordNotFound:                                                                                                                                                                                 
       Couldn't find OidcUser                                                                                                                                                                                      
     # ./app/controllers/oidc_users_controller.rb:46:in `destroy'                                                                                                                                                  
     # ./spec/requests/oidc_users_spec.rb:103:in `block (5 levels) in <top (required)>'                                                                                                                            
     # ./spec/requests/oidc_users_spec.rb:103:in `block (4 levels) in <top (required)>'     
  ```
[Triggered by this line of the tests](https://github.com/alphagov/account-api/pull/154/commits/3e90b2f995b4775e41ed87341a3b1cfe926016ed#diff-17d8a2354955c9bf68280e255dd8117c4c5baa2d51037d9b692390e6e6b9b354R103)

I tried removing the change test and just checking for the response code, but the tests still freak out at the existance of the error. And i tried `be_not_found` on the block (on the off chance the block syntax would let it bubble up to whaterver returned 404)... but no joy `be_not_found` does not support rspec block syntax.

When does action dispatch kick in here?
Id expected to be able to test this by just removing the clause and having my existing 404 test work... but perhaps not with request specs?

**Comment by huwd** [comment]
also woops forgot some docs.
[Added in a new commit](https://github.com/alphagov/account-api/pull/154/commits/0ab588941f5192078d9c9cee11c689c55d1214fb)

**Comment by huwd** [comment]
🤔 odd CI just showed a test faiure:
https://ci.integration.publishing.service.gov.uk/job/account-api/job/delete-users/4/console

But then when I rebuilt it worked, looks like we've got a very occasionaly flakey test on that PUT endpoint...

File: app/controllers/oidc_users_controller.rb (hunk 1 above)
Comment (aldavidson) [reply app/controllers/oidc_users_controller.rb:50]: Ah, request specs are not 'true' full-stack tests and are perhaps closer to controller specs in this respect. It's Rails trying to be helpful and intercepting the error to show you the full details in test mode, rather than giving a true reflection of what the user would see in production. See [this answer when this was raised as an issue in rspec-rails](https://github.com/rspec/rspec-rails/issues/2024#issuecomment-420809675).

There's two options, really:
1. test this with a feature spec instead, or
2. add a [workaround that tells Rails not to intercept the error](https://github.com/rspec/rspec-rails/issues/2024#issuecomment-420646336)

Comment (huwd) [reply app/controllers/oidc_users_controller.rb:50]: I suppose if we can trust that the interceptor works then the test could be:

```
expect { delete oidc_user_path(subject_identifier: subject_identifier), params: params, headers: headers }.to raise(ActiveRecord::RecordNotFound)
```

Then i can add a follow up feature test to check we get a 404?

Comment (huwd) [reply app/controllers/oidc_users_controller.rb:50]: [This is a great link](https://github.com/rspec/rspec-rails/issues/2024#issuecomment-420646336) by the way, thanks very much.

Super interetsing thread and to see an example of how-to!
I'm somewhat leaning that way now...

Comment (huwd) [reply app/controllers/oidc_users_controller.rb:50]: It works! I really like this,
It's explicit and contained. I'll note the reasons and link to this thread in a commit then force push.

Thanks Al! Great review

**Review by aldavidson — APPROVED** [review]
Nice! Glad to be of some help

File: docs/api.md [hunk 2]
```diff
@@ -225,6 +226,33 @@ Response:
...
+Request (with gds-api-adapters):
+
+```ruby
+GdsApi.account_api.delete_user(
+    govuk_account_session: "session-identifier",
+)
+```
+
+Response is status code only.
+
```
barrucadu, line 255: The endpoint is actually `DELETE /api/oidc-users/:subject_identifier`, and it doesn't take a `GOVUK-Account-Session` param

barrucadu: Looks like I forgot to add the `UPDATE /api/oidc-users/:subject_identifier` endpoint to this file when that was introduced, I'll open a PR for that

PR #158, opened 2021-07-26
Add new pact provider to build an oidc-user sub

huwd:
Pact tests in GDS API adaptors are [curretly][1] failing because there
is no is no provider_state in account-api with the expected name
`the-subject-identifier`.

This sets up a provider that will build a user record with that value
for the Pact to verify with.

[1]: https://github.com/alphagov/gds-api-adapters/pull/1098


barrucadu (commented):
Looks good, do the pact tests in https://github.com/alphagov/gds-api-adapters/pull/1098 pass when run against this branch?

barrucadu (approved):

PR #159, opened 2021-07-26
Move email subscription cancel checking logic into the model 

barrucadu:
This puts all of the email-alert-api communication in one place.  I've also moved an `update!` into an `ensure` block, to make it clearer that we always want it to run, and removed some superfluous `to_hash`es.

### DIFF
--- a/app/controllers/email_subscriptions_controller.rb
+++ b/app/controllers/email_subscriptions_controller.rb
@@ -4,17 +4,9 @@ class EmailSubscriptionsController < ApplicationController
   before_action :check_subscription_exists, only: %i[show destroy]
 
   def show
-    if email_subscription.email_alert_api_subscription_id
-      begin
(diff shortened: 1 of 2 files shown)


app/controllers/email_subscriptions_controller.rb
```diff
@@ -4,17 +4,9 @@ class EmailSubscriptionsController < ApplicationController
...
-        state = GdsApi.email_alert_api.get_subscription(email_subscription.email_alert_api_subscription_id)
-        if state.to_hash.dig("subscription", "ended_reason")
-          email_subscription.destroy!
-          head :not_found and return
-        end
-      rescue GdsApi::HTTPGone, GdsApi::HTTPNotFound
-        email_subscription.destroy!
-        head :not_found and return
-      end
+    if email_subscription.activated? && !email_subscription.check_if_still_active!
```
huwd, line 7: much more readable! Nice one

huwd (approved):

PR #166, opened 2021-07-30
Implement Transition Checker onboarding email

barrucadu:
When a user first signs up to Transition Checker notifications, we
want to send them an onboarding email.

This is part of the unbodging: the Account Manager is now sending a
much more generic email and so, on the GOV.UK side, we want to send
the more specific one.

This email will also be sent if the user signs up and only later uses
the Checker, whereas before they'd just have got the one
Checker-mentioning email regardless of whether they'd used the Checker
or not.

I thought about implementing the logic with some sort of check based
on attribute values, but decided it would be clearer (and less
fragile) to add a `has_received_transition_checker_onboarding_email`
field to the `OidcUser` model.  This new field defaults to `true`, so
we won't start sending any emails yet.  This lets us check the Notify
API keys and suchlike work in every environment, then we can deploy
another migration changing the default to `false` and begin sending
emails.

---

[Trello card](https://trello.com/c/aqRJqOKQ/914-send-brexit-checker-specific-welcome-email-when-someone-first-confirms-their-address)

### DIFF
--- a/app/models/email_subscription.rb
+++ b/app/models/email_subscription.rb
@@ -44,6 +44,8 @@ class EmailSubscription < ApplicationRecord
     )
 
     update!(email_alert_api_subscription_id: subscription.dig("subscription", "id"))
+
+    send_transition_checker_onboarding_email!(email)
   end
 
   def deactivate!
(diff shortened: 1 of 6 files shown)


app/models/email_subscription.rb
```diff
@@ -56,4 +58,18 @@ def deactivate!
   ensure
     update!(email_alert_api_subscription_id: nil)
   end
+
+  def send_transition_checker_onboarding_email!(email)
+    return unless email_alert_api_subscription_id
+    return unless name == "transition-checker-results"
```
huwd, line 64: Do we think we'll have others of these?
Wondering if we should put this in a const at the top or something and then do:
`return unless ALLOWED_NAMES.include?(name)`

Could always add that when we've got the second service though?

huwd (approved):
one, non-blocking comment.

app/models/email_subscription.rb
barrucadu: Yeah, I think we should leave any attempt at generalisation until we have another thing which has an onboarding email.  We can't predict what we'll need.

PR #174, opened 2021-08-05
Write up how we will do flash messages

barrucadu:
[Rendered](https://github.com/alphagov/account-api/blob/msw/flash/docs/adr/001-flash-messages.md)

---

[Trello card](https://trello.com/c/I0ulYzeX/946-implement-a-way-to-use-flash-messages-for-accounty-things)

### DIFF
--- a/docs/adr/001-flash-messages.md
+++ b/docs/adr/001-flash-messages.md
@@ -1,0 +1,121 @@
+Flash messages
+==============
+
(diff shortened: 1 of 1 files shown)


huwd (approved):
Sounds like a good idea,
and I like that this is written up in a simple ADR.

Nice one.

PR #185, opened 2021-08-25
Remove the `LocalAttribute` model & logic

barrucadu:
All `LocalAttribute` values have been migrated into `OidcUser`s in
production.  After this has been deployed we can deploy a migration to
drop the table.

This also refactors the `OidcUser#set_local_attributes` and
`OidcUser#get_local_attributes` methods.


aldavidson (approved):

barrucadu:
The migration code has been running for a while now, and I've confirmed that `LocalAttribute.where(migrated: false).count == 0`, so this is good to go.

PR #189, opened 2021-08-27
Remove unused development-mode branch

barrucadu:
Previously, we supported taking the account session value from the
request cookies when running in development mode.  This matches how
the frontend apps behave.  But in practice we don't make requests
directly to account-api itself from a browser, so this is unnecessary.


aldavidson (approved):
:+1:

PR #190, opened 2021-08-27
Store the ID token JWT in the session, and add an end-session endpoint

barrucadu:
This commit adds the original ID token value (a JWT) to the session, so that we can follow the [OIDC session management standard](https://openid.net/specs/openid-connect-session-1_0.html) Digital Identity have adopted.  The account manager doesn't use this standard (it has its own way of logging users out), so this is another bit of code with an `if using_digital_identity?` case - which isn't great, but is only temporary.

We'll need to add this endpoint to gds-api-adapters, then it can be used in the current [frontend SessionsController](https://github.com/alphagov/frontend/blob/5d7f08f7ad87033f3d602d378609aa6e74d720a0/app/controllers/sessions_controller.rb#L29-L38).

---

[Trello card](https://trello.com/c/ZvDGaba4/958-use-di-auth-in-integration-give-feedback-on-their-documentation)

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -24,6 +24,7 @@ class AuthenticationController < ApplicationController
     render json: {
       govuk_account_session: AccountSession.new(
         session_secret: Rails.application.secrets.session_secret,
+        id_token: details.fetch(:id_token_jwt),
         user_id: details.fetch(:id_token).sub,
         access_token: details.fetch(:access_token),
         refresh_token: details.fetch(:refresh_token),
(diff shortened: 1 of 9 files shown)


barrucadu:
This PR stores the original ID token, but DI are thinking through what happens if that expires.  If we have to request a new ID token, then I'll update this PR to *not* store it and to always fetch a new one when signing out (rather than have some time-based logic).

barrucadu:
Clarified that it is the original ID token we should use.

app/controllers/authentication_controller.rb
```diff
@@ -37,6 +38,24 @@ def callback
...
+        uri = oidc_client_class.new.end_session_endpoint
+        if id_token
+          querystring = Rack::Utils.build_nested_query(id_token_hint: id_token)
+          "#{uri}?#{querystring}"
+        else
+          uri
+        end
+      else
+        "#{Plek.find('account-manager')}/sign-out?continue=1"
+      end
```
aldavidson, line 54: Maybe pull some/all of this logic out into a separate private method? 
```suggestion
      if using_digital_identity?
        digital_identity_end_session_url
      else
        "#{Plek.find('account-manager')}/sign-out?continue=1"
      end
```

app/controllers/authentication_controller.rb
```diff
@@ -37,6 +38,24 @@ def callback
...
+
+    render json: { end_session_uri: end_session_uri }
+  end
+
 private
```
aldavidson, line 60:
```suggestion
  def digital_identity_end_session_url
   id_token = get_id_token_from_session
        uri = oidc_client_class.new.end_session_endpoint
        if id_token
          querystring = Rack::Utils.build_nested_query(id_token_hint: id_token)
          "#{uri}?#{querystring}"
        else
          uri
        end
  end
```

barrucadu:
I've split the URL generation into a separate method, and also removed a redundant `rescue OidcClient::OAuthFailure` (`AccountSession.deserialise` already rescues it).

aldavidson (approved):
Looks good :+1:

PR #194, opened 2021-09-03
Rename session_signing_key to session_secret

barrucadu:
It's not the key, it's the secret used to derive the key.

When the env var has been renamed in puppet, the old one can be
removed from config/secrets.yml.


richardTowers (approved):
Thanks for doing this! Code looks good to me. As far as I can tell this is a clean rename, so I haven't done any manual testing that the code still works. Happy to trust the tests if you are.

PR #201, opened 2021-09-15
Add a rake task to publish special routes

barrucadu:
Depends on https://github.com/alphagov/govuk-puppet/pull/11277

---

So far, these special routes have lived in special-route-publisher.
But we're going to publish the sorting hat page from account-api, as
it won't be a special route (it'll be a help page with some fancy
frontend).

When we have the content and frontend for the sorting hat page ready
to do, we'll add it to this `PublishingApiTasks` class, along with a
rake task.

Since it's potentially confusing having some accounts routes published
from one app and other accounts routes published by this app, I've
decided to move the existing special routes here.  When this commit is
merged, the config for these can be deleted from
https://github.com/alphagov/special-route-publisher/blob/master/data/special_routes.yaml

I've not moved over the `/transition-check` routes, as some of those
are account-related and others are not, and it's probably best to keep
all the transition checker routes together.  The only routes we should
publish from account-api are the ones which are related to core
account functionality.

---

[Trello card](https://trello.com/c/FbsTQJ0z/1008-build-sorting-hat-page-v1)


lib/publishing_api_tasks.rb
```diff
@@ -0,0 +1,74 @@
+require "gds_api/publishing_api/special_route_publisher"
+
+class PublishingApiTasks
+  SPECIAL_ROUTES = [
```
barrucadu, line 4: I've gone for defining the routes in the class here, but am open to moving this into a config file which gets read in.

aldavidson (approved):
looks good :+1:

PR #208, opened 2021-09-17
Add a rake task to check if a user exists

alex9smith:
Look up users by email address to see if they have an account. We have
to do this occasionally, so add a more convenient method than using the
Rails console.

[Trello card](https://trello.com/c/MoyeH1GU/1017-add-a-rake-task-to-check-if-an-account-exists)

### DIFF
--- a/spec/lib/support_spec.rb
+++ b/spec/lib/support_spec.rb
@@ -1,0 +1,31 @@
+Rails.application.load_tasks
+
+RSpec.describe "Support tasks" do
+  describe ":find_user" do
+    subject(:task) { Rake.application["support:find_user"] }
+
+    context "when there are no users" do
+      it "outputs the user doesn't exist" do
+        expect { task.execute({ email: "foo@example.gov.uk" }) }.to output("User 'foo@example.gov.uk' does not exist\n").to_stdout
+      end
+    end
+
+    context "when a user with email 'foo@example.gov.uk' exists" do
+      before do
+        FactoryBot.create(:oidc_user, email: "foo@example.gov.uk")
+      end
+
+      it "can find that user by email" do
+        expect { task.execute({ email: "foo@example.gov.uk" }) }.to output("User 'foo@example.gov.uk' exists\n").to_stdout
+      end
+
+      it "outputs other users don't exist" do
+        expect { task.execute({ email: "bar@example.com" }) }.to output("User 'bar@example.com' does not exist\n").to_stdout
+      end
+
+      it "performs a lowercase search for the user" do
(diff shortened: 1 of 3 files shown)


alex9smith:
Does this need to be documented anywhere? I had a quick look through the support part of the manual and didn't see anything similar

barrucadu:
Much of the stuff under https://team-manual.account.publishing.service.gov.uk/support/ will become irrelevant or need changing when we migrate stuff over to DI.  Since it's just our team who deals with account-related support stuff right now, let's not put the work in to update / review / rewrite our support docs before the DI migration (which should be late next month).

Also, checking which rake tasks are available, with `bundle exec rake -T`, is a pretty common practice amongst GOV.UK developers who do need to do some support task.  So this is discoverable.

spec/lib/support_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  describe ":find_user" do
+    subject(:task) { Rake.application["support:find_user"] }
+
+    context "when there are no users" do
+      it "outputs the user doesn't exist" do
+        expect { task.execute({ email: "foo@example.gov.uk" }) }.to output("User 'foo@example.gov.uk' does not exist\n").to_stdout
+      end
+    end
+
+    context "when a user with email 'foo@example.gov.uk' exists" do
```
barrucadu, line 13: Nitpick: I personally don't like specs with mutually exclusive `context`s like this.  Rather than

```ruby
describe "something to test" do
  context "when X is false" do
    ...
  end

  context "when X is true" do
    setup { ... }

    ...
  end
end
```

I'd usually remove the `context` around the `when X is false` case:

```ruby
describe "something to test" do
  ...

  context "when X is true" do
    before { ... }

    ...
  end
end
```

Though if both cases have some `before` I'd be more inclined to leave it.

But we don't have a standard for this, and other developers do it differently, so leave this if you want.

barrucadu (approved):
Looks good to me.  I'll have to merge this, so let me know if you want to change the specs or if I should just go ahead.

spec/lib/support_spec.rb
alex9smith: Thinking about it a bit more, do we even need the test for when there are no users? I'm not sure we gain anything worthwhile from it. If you agree, we could get rid of that test and the whole 'falsey' context block

barrucadu: Yeah, we already cover the case where the email address isn't found in other tests.  We can remove this.

alex9smith: Removed in 2558f32

alex9smith:
@barrucadu This is ready to merge now

PR #209, opened 2021-09-20
Drop `local_attributes` table

barrucadu:
The use of this was removed in #185.  This migration is, strictly
speaking, reversible: the data which was in this table was migrated
into the `oidc_users` table by #149.  So we could revert this and
deploy code to pull the data back out if need be.


alex9smith (approved):
LGTM

PR #210, opened 2021-09-20
Simplify levels of auth into an "MFA required?" flag

barrucadu:
Previously, we supported an arbitrary number of levels of authentication, so long as there's a total ordering on them (implemented as a pair of to / from integer functions).  But this is too general.  We've only ever needed two levels: one with MFA and one without.

We also expose level names in the public API, which isn't so great.  In principle, that means API consumers need to know about our levels, and what each one does, and adding or removing a level would be a breaking change.  In practice, it doesn't, because we only have two levels.

We'll need to change the level names when we migrate to DI auth.  So, rather than just keep the current situation and propagate that name change out to the consuming apps, let's rethink the API.  Change it to, rather than accept a level of auth as a parameter, instead just accept an MFA flag.  It's then up to the account-api to translate that into the appropriate parameter to pass to the auth service.

This PR refactors the permission model to just use the two levels of MFA / not-MFA, and adds the new MFA parameter (for the auth endpoint) and response field (for the user endpoint).  The old level-of-authentication parameter & field remain until we update gds-api-adapters and the consuming apps.

So there will be a few follow-up PRs for this:

1. One updating gds-api-adapters
2. Then, one updating finder-frontend (the only app to pass a level of authentication parameter)
3. Then, one removing the old API bits from account-api

At some point, maybe today, there'll be a fourth PR implementing the DI levels.

---

[Trello card](https://trello.com/c/ZvDGaba4/958-use-di-auth-in-integration-give-feedback-on-their-documentation)


app/controllers/authentication_controller.rb
```diff
@@ -1,11 +1,21 @@
 class AuthenticationController < ApplicationController
   include DigitalIdentityHelper
 
+  DEFAULT_LEVEL_OF_AUTHENTICATION = "level0".freeze
```
huwd, line 4: So this is just for a transition phase right?
Like eventually we'll remove this and just use the flag, but we're leaving this to ease transition?

huwd (approved):

app/controllers/authentication_controller.rb
barrucadu: Yes, some more level-of-auth terminology goes away in [my branch implementing the DI thing](https://github.com/alphagov/account-api/compare/msw/simplify-levels-of-auth-2...msw/vtr)

PR #215, opened 2021-09-23
Implement Vectors of Trust (RFC 8485) for DI auth

barrucadu:
This is a draft as I'm waiting for DI to confirm that the `vtr` query param is the right way to pass the values, as the RFC describes a different way.

---

With the account manager we have numbered levels (these could be
considered "scalars of trust"), but DI are aligning to an actual
standard.

The `vtr` parameter is a JSON array of acceptable vectors of
trust ("vot"s).  A vot is made up of one or more vector component
values, separated by dots.

The vector component values are:

- `Cl` = username & password
- `Cm` = MFA

So the vot `"Cl"` means "authenticate with a username & password",
whereas the vot `"Cm"` means "authenticate with MFA" (and just MFA -
not all vots are valid).

Dots denote conjunction.  So the vot `"Cl.Cm"` means "authenticate
with username & password & MFA".  This corresponds to our current
`level1`.

Multiple vots in the vtr denote disjunction.  So the vtr `["Cl",
"Cl.Cm"]` means "authenticate with username & password, or with
username & password & MFA".  This corresponds to our current `level0`.

The actual vot uses is returned in the `vot` claim in the ID token.
So we can check that to see if the user authenticated with MFA or not.

See also

- [The DI ADR](https://github.com/alphagov/digital-identity-architecture/blob/main/authentication/adr/0005-mechanism-for-selecting-authn.md)
- [Section 3.2 of RFC 8485](https://datatracker.ietf.org/doc/html/rfc8485#section-3) (retrieving `vot` from the ID token)
- [Section 4.1 of RFC 8485](https://datatracker.ietf.org/doc/html/rfc8485#section-4.1) (specifying `vtr` value when making requests)

---

[Trello card](https://trello.com/c/ZvDGaba4/958-implement-vectors-of-trust-di-levels-of-authentication)


app/lib/oidc_client/account_manager.rb
```diff
@@ -8,6 +8,27 @@ def auth_uri(auth_request, mfa: false)
...
+    response = get_ephemeral_state(
+      access_token: tokens[:access_token],
+      refresh_token: tokens[:refresh_token],
+    )
+
+    tokens.merge(
+      access_token: response.fetch(:access_token),
+      refresh_token: response.fetch(:refresh_token),
+      mfa: response.fetch(:result).fetch("level_of_authentication") == "level1",
+      ga_session_id: response.fetch(:result)["_ga"],
```
huwd, line 27: Do these two also need a :
```
# TODO: remove `ga_client_id` and `cookie_consent` after we switch
# to DI, we're using a different approach there which doesn't go
# via the account-api
```
?

huwd (approved):
LGTM, one thought on TODO comments so we don't forget

app/lib/oidc_client/account_manager.rb
barrucadu: No, because we'll just remove this class when the account manager goes away.

barrucadu:
Getting an error response from DI with this branch, trying to authenticate redirects back with

```
error=invalid_request
error_description=Request+vtr+not+valid
```

PR #217, opened 2021-09-27
ADR 003: Cookie consent

barrucadu:
This is a summary of the approach to shared consent we've agreed with
the DI auth team.

---

[Trello card](https://trello.com/c/xr7VyhgV/1034-write-an-adr-for-cookie-consent-cross-domain-analytics)


huwd (approved):
clear and well written.
Nice work!

PR #223, opened 2021-09-30
Populate `legacy_sub`, add an index, and partially implement user migration

barrucadu:
After writing the commit message of b193159, I realised that we don't
actually want this column to be non-nullable.  Because,
post-migration, we will want to create users who have a `sub` but not
a `legacy_sub`.

Usefully, SQL allows multiple NULL values in a nullable field which
has a unique index.  I imagine this is because NULLs are incomparable
in SQL (both `NULL = NULL` and `NULL != NULL` are false).

The change to `OidcUsersController` will let us update or delete users
by legacy subject identifier.  When updating a user, we'll also save the
new subject identifier.  This will gradually migrate our data as people
update their account.

We will also need to implement user migration in the `AccountSession`,
but there are some prerequisites (noted in a comment).

### DIFF
--- a/db/migrate/20210930105948_populate_legacy_sub_and_create_index.rb
+++ b/db/migrate/20210930105948_populate_legacy_sub_and_create_index.rb
@@ -1,0 +1,21 @@
+class PopulateLegacySubAndCreateIndex < ActiveRecord::Migration[6.1]
+  def up
+    OidcUser.update_all("legacy_sub = sub")
+
+    # The SQL standard (and, more relevantly to us, Postgres) allows a
+    # nullable field with a unique index to have multiple NULL values.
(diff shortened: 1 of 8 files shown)


db/migrate/20210930105948_populate_legacy_sub_and_create_index.rb
```diff
@@ -0,0 +1,21 @@
+class PopulateLegacySubAndCreateIndex < ActiveRecord::Migration[6.1]
+  def up
+    OidcUser.update_all("legacy_sub = sub")
+
+    # The SQL standard (and, more relevantly to us, Postgres) allows a
```
huwd, line 5: This is really useful, nice one.

huwd (approved):

PR #228, opened 2021-10-07
Migrate users as they authenticate

barrucadu:
When we authenticate a user through Digital Identity, check if we have
a record in our database for them.  If we do, all is well.  If not,
we're in one of two cases;

- This user hasn't logged in to GOV.UK at all before
- This user hasn't logged in to GOV.UK since the migration

We can distinguish the two cases by calling the UserInfo endpoint, and
checking the `govuk-account` claim.  If present, this claim holds the
pre-migration subject identifier, which we can use to look up and
migrate the user in our database.

---

[Trello card](https://trello.com/c/eRBL2kEZ/1050-implement-migration-of-subject-identifiers)

### DIFF
--- a/app/lib/account_session.rb
+++ b/app/lib/account_session.rb
@@ -43,7 +43,10 @@ class AccountSession
   end
 
   def user
-    @user ||= OidcUser.find_or_create_by_sub!(user_id)
+    @user ||= OidcUser.find_or_create_by_sub!(
+      user_id,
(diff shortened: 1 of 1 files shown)


barrucadu:
@alex9smith I realised this PR was slightly wrong, see b8e9d3fe74ebd87a8feab787a71195a2aeabf4c9

alex9smith:
Good spot!

PR #251, opened 2021-10-26
Record when a user is destroyed, and end sessions for destroyed users

barrucadu:
See commit messages for details.

---

[Trello card](https://trello.com/c/goXbVvpM/1113-record-when-a-user-is-deleted)

### DIFF
--- a/app/lib/account_session.rb
+++ b/app/lib/account_session.rb
@@ -9,6 +9,8 @@ class AccountSession
 
   class MissingCachedAttribute < ReauthenticateUserError; end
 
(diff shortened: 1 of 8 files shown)


huwd (commented):
Left a few thoughts in slack.
Code looks great - just want to check in on a few bits about the approach [here](https://gds.slack.com/archives/C011Y5SAY3U/p1635855023026500?thread_ts=1635850403.025800&cid=C011Y5SAY3U)

huwd (approved):

PR #261, opened 2021-11-01
Add endpoint to match a user by email address & add an index on the field

barrucadu:
This will be used to implement the sniffer page in
email-alert-frontend, where we want to examine the email address the
user has just entered and do one of three things:

- Log the user in to email-alert-frontend if they're logged into the
right GOV.UK account; or

- Tell the user to sign in, if a GOV.UK account exists with the email
address they entered; or

- Fall back to the magic link method if there is no GOV.UK account for
that email address.

email-alert-frontend will use this endpoint like so:

    def match_user_by_email(email)
      response = GdsApi.match_user_by_email(
        email: @address,
        govuk_account_session: account_session_header,
      )

      if response["match"]
        :ok
      else
        :must_reauthenticate
      end
    rescue GdsApi::HTTPNotFound
      :no_account
    end

The alternative would be to add an endpoint to find a user by email
address, and then for email-alert-frontend to do something like this:

    def match_user_by_email(email)
      response = GdsApi.account_api.get_attributes(
        attributes: %w[email],
        govuk_account_session: account_session_header,
      )

      return :ok if response["values"]["email"] == email

      GdsApi.account_api.find_user_by_email(email: email)
      :must_reauthenticate
    rescue GdsApi::HTTPNotFound
      :no_account
    end

which would be slower because it makes two API calls rather than one.

---

[Trello card](https://trello.com/c/5ZoujfPL/1115-implement-email-management-account-sniffer-logic)

### DIFF
--- a/app/controllers/concerns/authenticated_api_concern.rb
+++ b/app/controllers/concerns/authenticated_api_concern.rb
@@ -1,10 +1,12 @@
 module AuthenticatedApiConcern
   extend ActiveSupport::Concern
 
+  HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION".freeze
+
   included do
     before_action do
       @govuk_account_session = AccountSession.deserialise(
-        encoded_session: request.headers["HTTP_GOVUK_ACCOUNT_SESSION"],
+        encoded_session: request.headers[HEADER_NAME],
         session_secret: Rails.application.secrets.session_secret,
       )
 
--- a/app/controllers/match_user_by_email_controller.rb
+++ b/app/controllers/match_user_by_email_controller.rb
@@ -1,0 +1,32 @@
+class MatchUserByEmailController < ApplicationController
+  before_action :fetch_session_if_present
+
+  def show
+    email = params.fetch(:email).downcase
+
+    if @govuk_account_session&.user&.email == email
+      render json: { match: true }
+      return
+    end
+
+    if OidcUser.where(email: email).exists?
+      render json: { match: false }
+    else
+      head :not_found
+    end
+  end
+
+private
+
+  def fetch_session_if_present
+    encoded_session = request.headers[AuthenticatedApiConcern::HEADER_NAME]
+    return unless encoded_session
+
+    @govuk_account_session = AccountSession.deserialise(
+      encoded_session: encoded_session,
+      session_secret: Rails.application.secrets.session_secret,
+    )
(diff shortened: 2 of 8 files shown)


app/controllers/concerns/authenticated_api_concern.rb
```diff
@@ -1,10 +1,12 @@
 module AuthenticatedApiConcern
   extend ActiveSupport::Concern
 
+  HEADER_NAME = "HTTP_GOVUK_ACCOUNT_SESSION".freeze
```
huwd, line 4: I generally like stuff like this broken out - but why now?
Was it just something you spotted on the way past?

huwd: Oh cool, [there's my answer](https://github.com/alphagov/account-api/pull/261/commits/6400a2ebcd34b53bafbaa4b5d8522ad1620b872e#diff-7e6894647a98a35c83ab744086a6f542f6022c2837ec1ca905ce629ec22cd5c3R22)

app/controllers/match_user_by_email_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    if @govuk_account_session&.user&.email == email
+      render json: { match: true }
+      return
+    end
+
+    if OidcUser.where(email: email).exists?
+      render json: { match: false }
+    else
+      head :not_found
```
huwd, line 15: So a logged in user could use this endpoint to inferr if there was another account by email, right?
We're not limiting checks to just the account that is making the query?

I guess you could do the same with our login form, and at least this way we'd have a logged in user making this request.

Should we consider rate limiting this at all?

barrucadu: email-alert-frontend has a rate limiter: https://github.com/alphagov/email-alert-frontend/blob/main/app/services/verify_subscriber_email_service.rb - we could re-use that.

And yes, this intentionally doesn't require a logged in user, because for the Taxon / Topic Updates journey we *don't* have a logged-in user at the point where we look up an email address, and I wanted this endpoint to work for both (rather than have two very-similar-but-slightly-different endpoints)

barrucadu:
I've changed this to ignore an invalid session value rather than throw an error - it's more useful in email-alert-frontend to get either `match: false` or a 404 than a 401, and `match: false` prompts reauthentication anyway.

huwd (approved):

PR #264, opened 2021-11-03
Add rake task to move unmigrated users to a new table

barrucadu:
There are ~7000 users who didn't confirm their email address before
the cut-off date, and so who didn't get migrated.  9 of those have
since re-registered, which means we have 9 `OidcUser` records with
duplicate email addresses.

This causes a problem, because email addresses are otherwise unique.
So, let's move these users to another table.  Eventually we will
delete them, but that won't happen until after the brexit checker goes
away at least.

---

[Trello card](https://trello.com/c/eIgHkyyw/1123-move-unmigrated-users-to-a-different-database-table)

### DIFF
--- a/lib/tasks/migration.rake
+++ b/lib/tasks/migration.rake
@@ -1,0 +1,20 @@
+namespace :migration do
+  desc "Move unmigrated users to the unmigrated_oidc_users table"
+  task move_unmigrated_users_to_new_table: :environment do
+    users_to_migrate = OidcUser.where(email_verified: false).where("created_at < ?", Date.new(2021, 10, 28))
+    puts "migrating #{users_to_migrate.count}"
+
+    users_to_migrate.each do |user|
+      UnmigratedOidcUser.create!(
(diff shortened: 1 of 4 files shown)


lib/tasks/migration.rake
```diff
@@ -0,0 +1,20 @@
...
+    users_to_migrate.each do |user|
+      UnmigratedOidcUser.create!(
+        sub: user.sub,
+        email: user.email,
+        email_verified: user.email_verified,
+        has_unconfirmed_email: user.has_unconfirmed_email,
+        transition_checker_state: user.transition_checker_state,
+        created_at: user.created_at,
+        updated_at: user.updated_at,
+      )
```
alex9smith, line 16: Do you need a `.save` on the new unmigrated user?

alex9smith: Oh wait no, you've done `create!` which also saves

alex9smith (approved):

PR #268, opened 2021-11-04
Add rake task to merge pre- & post-migration users

barrucadu:
This is to handle the case where a legacy_subject_id isn't being
returned from the userinfo, and so we're trying to create a new
OidcUser record with the same email address as an existing,
pre-migration, one.

We have one such case so far (brought to light by the new unique index
on the email field).  I've asked DI to check the record for that user,
as they should have been migrated across.  But, so that the user can
log in, we need to merge their two records.

The task will only merge users where:

- the post-migration user has no legacy_sub or email.
- the pre-migration user has the same sub and legacy_sub.
- both users were created on the right sides of the migration.

---

[Trello card](https://trello.com/c/GYLhX263/1125-investigate-improperly-migrated-user)

### DIFF
--- a/lib/tasks/migration.rake
+++ b/lib/tasks/migration.rake
@@ -17,4 +17,37 @@ namespace :migration do
       user.destroy!
     end
   end
+
+  desc <<~DESC
+    Merge a pre-migration and an improperly-created post-migration user
+
+    This is to handle the case where a legacy_subject_id isn't being
+    returned from the userinfo, and so we're trying to create a new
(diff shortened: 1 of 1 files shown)


barrucadu:
Upon further investigation, we had a bug on our end where users could confirm their email addresses after the cut-off point, as we didn't take down the route which does that.

This affects 179 users.  Some of these have brexit checker results saved, so we don't really want to delete them.  Instead we can just use this task to merge them as they come up.

huwd (approved):

PR #269, opened 2021-11-04
Store exceptions from updating users to a new table

barrucadu:
These exceptions can be raised due to non-unique email addresses.
Such an exception indicates a bug, as email addresses should be
unique, but we don't want real user data getting into Sentry.


huwd (approved):

barrucadu:
Oh actually, we don't want to rescue the `CapturedSensitiveException` as that stops it getting to Sentry.  One moment...

PR #273, opened 2021-11-05
Make check-email-subscription endpoint also take a base_path

barrucadu:
If a base_path is given, we match against the URL field of the
subscriber list.  This is so we can find a subscription for a specific
page, without needing the frontend app to query email-alert-api for
the slug first.

I've also added tests for the case where a user has active
subscriptions which don't match, as that wasn't covered previously.

---

[Trello card](https://trello.com/c/oOpGMmxF/1126-put-the-single-page-signup-button-on-some-pages)


alex9smith (approved):
Nice!

PR #278, opened 2021-11-10
Add single-page button HTML to the API response 

barrucadu:
This is so that the javascript can just replace the existing button,
rather than trying to do some more fancy DOM manipulation.

---

To try this out locally, you'll also need #277 

---

[Trello card](https://trello.com/c/oOpGMmxF/1126-put-the-single-page-signup-button-on-some-pages)

### DIFF
--- a/app/controllers/check_email_subscription_controller.rb
+++ b/app/controllers/check_email_subscription_controller.rb
@@ -62,6 +62,14 @@ private
       base_path: @base_path,
       topic_slug: @topic_slug,
       active: active,
+      button_html: render_button_component(active),
     }.compact
+  end
+
+  def render_button_component(active)
+    ActionController::Base.render(
+      partial: "govuk_publishing_components/components/single_page_notification_button",
+      locals: { base_path: @base_path, already_subscribed: active },
+    ).presence
   end
 end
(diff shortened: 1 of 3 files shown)


alex9smith:
This looks good to me, but I'm getting `{"message":"No bearer token was provided"}` when I try and test locally with curl (I have pulled in the cookie middleware from the linked PR).

I wasn't expecting that based on the docs - am I doing something wrong?

barrucadu:
Is your cookie (or header) valid?

I think the `include GDS::SSO::ControllerMethods` in the `ApplicationController` is intercepting 401 responses and giving that message.  I'm planning a refactoring PR which adds some different superclasses, since it seems that skipping the `authorise_sso_user!` action isn't enough to fully remove the gds-sso stuff in this case.

alex9smith:
It should be, I'm logged in to `http://frontend.dev.gov.uk/account/home` and getting the session cookie from there. 

Your explanation seems reasonable and the tests cover the invalid and unauthenticated states so I'm going to approve as to not block the last few bits of work, and I'll work out what's going on on my local setup separately.

alex9smith (approved):

PR #279, opened 2021-11-11
Refactor: move internal & personalisation controllers into separate modules, & only pull in cookies middleware in development mode

barrucadu:
See commits for details

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,4 @@
 class ApplicationController < ActionController::API
-  include GDS::SSO::ControllerMethods
-
   class CapturedSensitiveException < StandardError
     attr_reader :captured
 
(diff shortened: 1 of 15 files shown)


alex9smith:
Interestingly, I noticed while testing that the check email endpoint returns a rendered HTML 401 page if you're not logged in. Presumably that's because the request is going through the full cache -> router -> app stack and something is intercepting the 401 status? I'm not sure there's anything we can do about that. 

I also had a look through the warden docs to see if there was a `before_action` we could skip or add to our controllers to stop it intercepting, but I couldn't see anything quite like that.

As an aside, what are gds-sso and warden doing in the account-api? Are they for app to app authentication? We have our own authentication and session management for end user sessions.

alex9smith (approved):

barrucadu:
> Presumably that's because the request is going through the full cache -> router -> app stack and something is intercepting the 401 status? I'm not sure there's anything we can do about that.

Yes, we have something in nginx (or maybe router) which catches errors from frontend apps and serves a static template.  This is kind of confusing and on the GOV.UK tech debt board because it's not what you'd expect.

> As an aside, what are gds-sso and warden doing in the account-api? Are they for app to app authentication?

It's for authenticating app-to-app communication with bearer tokens.

PR #304, opened 2021-12-10
Use `extra`, not `tags`, for captured sensitive exception IDs

barrucadu:
`tags` is a way of filtering events.  When looking at a group of
events, you can click the "tags" panel, and then click a tag value to
filter to only those events with the same tag.

`extra` is a way of adding additional information to a single event.
When looking at a group of events, each has an "Additional Data"
section (beneath the "Operating System" and "Runtime" sections).  No
changing of panels needed, it's right there.

So change to `extra` here as it's more correct (and more convenient)
for what we're trying to do: attach unique data to each event.

---

```ruby
GovukError.notify("Michael's Test Error", { tags: { some_tag_here: "tag-value" } })
```

Screenshot 2021-12-10 at 12 06 37

```ruby
GovukError.notify("Michael's Test Error", { extra: { extra_data_here: "extra-data-value" } })
```

Screenshot 2021-12-10 at 12 06 59

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,27 +1,25 @@
 class ApplicationController < ActionController::API
-  class CapturedSensitiveException < StandardError
-    attr_reader :captured
+  class CapturedSensitiveException < StandardError; end
 
-    def initialize(captured)
-      super()
-      @captured = captured
-    end
-  end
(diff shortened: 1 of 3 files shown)


alex9smith (approved):
Nice!

PR #325, opened 2022-01-05
Remove use of refresh tokens entirely

barrucadu:
Our pre-DI auth prototype service issued refresh tokens.  We used
these to fetch the UserInfo response when a user tried to use an
attribute which we don't have cached for them.

DI do not do refresh tokens, so we made the minimal change needed to
our code: we added support for there *not* being a refresh
token (rather than assuming one always exists) and also made the
attribute-caching happen as soon as the access token is fetched, not
when the attribute is needed.  But we left the ability to use refresh
tokens in place.

DI have spotted us using invalid access tokens on occasion, and have
suggested that we might be re-using tokens.  This is kind of weird as
we now only use the access token to fetch the userinfo in the auth
callback: there are two places where we do that, but we share the
userinfo response if it gets fetched in the first place.  So we
shouldn't try to use the same access token twice.

I suspect that the change we made to support DI was not quite right.

If a userinfo request fails due to a 4xx error (which should ideally
not happen anyway), we'll call `OidcClient#refresh_client_tokens` and
retry the request.  We don't have anything to skip that if the refresh
token is missing, so my hunch is that that function just returns the
access token the client already has, and so one token is getting used
a second time.

Since we don't need refresh tokens in production, we can simplify the
OidcClient by stripping them out entirely.  If access tokens are still
being reused, this wasn't the real problem and we'll need to do some
more digging.  But even if that is the case, this change is positive.

---

[Trello card](https://trello.com/c/TbmyNJn9/1193-look-into-access-code-re-use-bug)

### DIFF
--- a/app/lib/oidc_client.rb
+++ b/app/lib/oidc_client.rb
@@ -40,13 +40,9 @@ class OidcClient
     end
 
     unless OidcUser.where(sub: tokens[:id_token].sub).exists?
-      response = userinfo(access_token: tokens[:access_token], refresh_token: tokens[:refresh_token])
-      OidcUser.find_or_create_by_sub!(tokens[:id_token].sub, legacy_sub: response.dig(:result, "legacy_subject_id"))
-      tokens.merge!(
-        access_token: response.fetch(:access_token),
-        refresh_token: response[:refresh_token],
-        userinfo: response[:result],
-      )
+      response = userinfo(access_token: tokens[:access_token])
+      OidcUser.find_or_create_by_sub!(tokens[:id_token].sub, legacy_sub: response["legacy_subject_id"])
+      tokens.merge!(userinfo: response)
     end
 
     tokens.merge(
@@ -83,7 +79,6 @@ class OidcClient
 
     {
       access_token: response[:access_token],
-      refresh_token: response[:refresh_token],
       id_token_jwt: id_token_jwt,
       id_token: id_token,
     }.compact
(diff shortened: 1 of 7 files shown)


app/lib/oidc_client.rb
```diff
@@ -83,65 +79,45 @@ def tokens!(oidc_nonce: nil)
...
   end
 
 private
 
-  class RefreshAndRetry < StandardError; end
+  class Retry < StandardError; end
 
-  class OnlyRetry < StandardError; end
-
-  REFRESH_AND_RETRY_STATUSES = [401, 403].freeze
```
huwd, line 117: I'm a little unclear what happens in a 401 and 403 case now.
previously we would have retried... once (?), then hit `MAX_OAUTH_RETRIES` and then raised `OAuthFailure`...

Presumably that just renderes an error page (is that the "something has gone wrong" page? or something more descriptive?), or do we just boot a user back to the login form?

So now for 401 and 403 we just jump straight to those pages, right?
No retries because those are definative "we see you and you're not allowed further" statuses.

Do we think any error pages we might serve in response to these are descriptive enough?
I can't quite see how this'd look from a user's perspective if they ended up on this branch of code?

barrucadu: This will be returned as a 401 from account-api to the consuming app.  Since the retry would have failed anyway (as there's no refresh token to use to get a new access token), there should be no change in response to other apps.

huwd (approved):

PR #379, opened 2022-03-17
Add a worker to delete old `SensitiveExceptions`

alex9smith:
We use `SensitiveExceptions` to store error messages which may contain
personal data so we can report the error in Sentry or Kibana but only
store the data and error message in our database.

We're also planning to use this mechanism to capture error responses
from the Auth team's OIDC API. These may contain authorisation codes
and access tokens which would allow access to a user's account.

We need a mechanism which automatically deletes these database records
once they're no longer useful to us. We should be getting a Sentry alert
and investigation quickly once an exception is raised, so we don't
need to keep them around for very long.

`AuthRequest`s have a very similar mechanism for cleaning up old entries
so reusing that makes sense.

We chose 14 days as the retention period as this matches how long GOV.UK
stores application logs in Kibana. There's no point in keeping this data
longer than the application logs as it's not very useful on its own.

---

[Trello](https://trello.com/c/15WqAeVo/1310-add-a-worker-to-account-api-that-deletes-old-sensitiveexceptions)

### DIFF
--- a/app/models/sensitive_exception.rb
+++ b/app/models/sensitive_exception.rb
@@ -1,2 +1,4 @@
 class SensitiveException < ApplicationRecord
+  EXPIRATION_AGE = 14.days
+  scope :expired, -> { where("created_at < ?", EXPIRATION_AGE.ago) }
 end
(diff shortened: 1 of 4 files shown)


huwd (commented):
PRs like this remind me why I like ruby and rails so much.
So clean and readable.

Nice one, great commit message too.

huwd (approved):

PR #405, opened 2022-05-18
Logout token

huwd:
[Trello](https://trello.com/c/p3328fd0/1346-create-a-new-endpoint-to-support-auth)

We're implimenting OIDC backchannel logout, to recieve messages from digital identity
indicating the uses wishes to have their session removed everywhere.

The initial stage of this is to:

- give `account-api` the awarenss of a LogoutToken
- Validate and verify contents [following the specification](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation)
- Ensure the JWT can be decoded correctly and the signature verified

This PR adds a Plain old ruby class to do just that.

### DIFF
--- a/app/lib/logout_token.rb
+++ b/app/lib/logout_token.rb
@@ -121,4 +121,10 @@ private
     Redis.current.set("logout-token/#{jti}", "OK")
     Redis.current.expire("logout-token/#{jti}", 2.minutes)
   end
+
+  class << self
+    def decode(jwt_string, key)
+      new JSON::JWT.decode jwt_string, key
+    end
+  end
 end
--- a/app/lib/logout_token.rb
+++ b/app/lib/logout_token.rb
@@ -1,0 +1,124 @@
+require "attr_required"
+require "attr_optional"
+
+class LogoutToken
+  include AttrOptional
+  include AttrRequired
+  class InvalidToken < RuntimeError; end
+  class ExpiredToken < InvalidToken; end
+  class InvalidIssuer < InvalidToken; end
+  class InvalidAudience < InvalidToken; end
+  class InvalidIssuedAt < InvalidToken; end
+  class InvalidEvent < InvalidToken; end
+  class InvalidBackchanelLogoutEvent < InvalidToken; end
+  class InvalidIdentifiers < InvalidToken; end
+  class TokenRecentlyUsed < InvalidToken; end
+  class ProhibitedNonse < InvalidToken; end
+
+  BACK_CHANNEL_EVENT_NAME = "http://schemas.openid.net/event/backchannel-logout".freeze
+
+  attr_required :iss, :aud, :iat, :jti, :events
+  attr_optional :sub, :sid, :auth_time
+  attr_accessor :access_token, :code, :state
+  alias_method :subject, :sub
+  alias_method :subject=, :sub=
+
+  NON_STRING_ATTRIBUTES = %i[aud exp iat auth_time sub_jwk events].freeze
+
+  def initialize(attributes = {})
+    attributes_to_methods(attributes)
+    apply_attribute_values_types(all_attributes)
+    validate_prohibited_attribtues(attributes)
+  end
+
+  def verify!(expected = {})
+    validates_issuer(expected)
+    validates_audience(expected)
+    validates_issued_at_time
+    validates_session_and_or_user_id_presence
+    validates_events(expected)
+    validate_jti_not_recently_used
+
+    true
+  end
+
+private
+
+  def attributes_to_methods(attributes)
+    all_attributes.each do |attr|
+      send :"#{attr}=", attributes[attr]
+    end
+    attr_missing!
+  end
+
+  def all_attributes
+    self.class.required_attributes + self.class.optional_attributes
+  end
+
+  def apply_attribute_values_types(all_attributes)
+    (all_attributes - NON_STRING_ATTRIBUTES).each do |key|
+      send "#{key}=", send(key).try(:to_s)
+    end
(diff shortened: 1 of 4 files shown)


app/lib/logout_token.rb
```diff
@@ -0,0 +1,124 @@
...
+  attr_accessor :access_token, :code, :state
+  alias_method :subject, :sub
+  alias_method :subject=, :sub=
+
+  NON_STRING_ATTRIBUTES = %i[aud exp iat auth_time sub_jwk events].freeze
+
+  def initialize(attributes = {})
+    attributes_to_methods(attributes)
+    apply_attribute_values_types(all_attributes)
+    validate_prohibited_attribtues(attributes)
```
alex9smith, line 31: Interesting that we're validating the `nonse` here but all the other attributes only get checked when we call `verify!` What's the reason for that?

app/lib/logout_token.rb
```diff
@@ -0,0 +1,124 @@
...
+    unless sub || sid
+      raise InvalidIdentifiers, "Invalid Logout token: Must contain either SID, SUB or both"
+    end
+  end
+
+  # Validates the logout token has a correctly formatted events claim
+  def validates_events(_expected = {})
+    raise InvalidBackchanelLogoutEvent, "Invalid Logout token: Events is not a hash" unless events.is_a? Hash
+    unless events.keys.include?(BACK_CHANNEL_EVENT_NAME)
+      raise InvalidBackchanelLogoutEvent, "Invalid Logout token: Events is not a hash"
```
alex9smith, line 99: Super nit pick - should this be a different error message? `events` is a hash if we've got this far, but it's missing the key we're looking for

app/lib/logout_token.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  # Verify that we've not recently recieved another request with a jti
+  def validate_jti_not_recently_used
+    raise TokenRecentlyUsed if is_recent_jti?(jti)
+
+    record_jti_as_recently_verified(jti)
+  end
+
+  def is_recent_jti?(jti)
+    return true if Redis.current.get("logout-token/#{jti}") == "OK"
```
alex9smith, line 115: This is nice!

alex9smith (approved):
Looks good! Couple of points which are more for my own curiosity than anything else

app/lib/logout_token.rb
huwd: Oh great, yeah not a nitpick at all, I should have spotted that!

app/lib/logout_token.rb
huwd: So I'll just play back out chat here for posteirty. It's a good question.
There's an OK reason for this and a bit of a naff reason.

## OK reason: Validating attributes vs Validating values

The OK reason is that there is a split in validation here:
I based the `logout_token.rb` on the [id_token.rb](https://github.com/nov/openid_connect/blob/master/lib/openid_connect/response_object/id_token.rb) as it is in[ our dependency](https://github.com/alphagov/account-api/blob/main/Gemfile#L20).

The author there created their [own gem](https://github.com/nov/attr_required) and used it in their [openid_connect id_token](https://github.com/nov/openid_connect/blob/master/lib/openid_connect/response_object/id_token.rb#L10-L11). The theory is, don't initalise the object without throwing an error if the core named parameters are not present.

I've re-used that here.

So then why validate prohibited attribtues here? The[ back_channel_logout spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation) has an odd bit:

> Verify that the Logout Token does not contain a nonce Claim. 

So we've got something `attr_required` doesn't cater for here as a prohibited attribute, so I rolled my own.

In general I'm seeing the difference as:
- In initalize check the attributes are present / as expected enough to be happy to continue
- verify! check the values and confirm if it's valid

I'm not entirely sure I like this pattern, GOV.UK folks always instilled in me that we should keep very little going on in initalize... Here we've got quite a lot.

## Rubbish Reason: So why keep it this way?

I kept it this way becuase I wanted to keep things looking roughly like the openid_connect gem. This is a rubbish reason because I'm favouring the style / format of a dependency instead of how we've written things elsewhere in this codebase... which feels a bit... eww.

But it made sense when writing the logout token as [the spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation) mentions a bunch of things that should be verified in the same way we do with an ID token. And I'm still hoping the discussion will continue on an issue I raised in [the dependency](https://github.com/nov/openid_connect/issues/67). In an ideal world, I could contribute some portion of this code back as a PR, and then we could outsource this token code to a module that comes with the ID token checks.

This doesn't seem like the most compelling of arguments, but it's where I ended up, and I don't think the cost of a somewhat busy initaliser is too bad, and I think the code is legible enough to merge.

Open to 🧠 if you disagree!

app/lib/logout_token.rb
huwd: [Fixed](https://github.com/alphagov/account-api/pull/405/commits/bec4eb258546849b571118cf7dfdbca5b73c3f69#diff-dca85ccfe8b7cfd1a78785e20834d6867814c2c3ffc788e470723b3e64dca00eR99): Cheers for spotting that

huwd:
Gonna ask for a re-review to check that [this answer](https://github.com/alphagov/account-api/pull/405#discussion_r879276810) reflects our Friday chat and your happy to continue on that basis.

If so, I'll merge and we can rebase the next one!

alex9smith (approved):

app/lib/logout_token.rb
alex9smith: No, that sounds good to me 👍🏻

PR #407, opened 2022-05-19
Extend OidcClient to add a logout_token endpoint

huwd:
[Trello](https://trello.com/c/p3328fd0/1346-create-a-new-endpoint-to-support-auth)
[Merge after](https://github.com/alphagov/account-api/pull/405)

This will recieve an encoded JWT:
- Verify the signature (or throw an error)
- Then ask the LogoutToken class to validate the payload

Assuming all is well there it'll return a hash with a request_time, the
token as a JWT and the token itself. Mostly trying to keep pairty with
the id_token inclusion in the tokens! method. But I can see how those'd
be useful for debugging too.

Testing this is a bit of a pain. Discovery reaches out to the
.well-known endpoints of the Identity Provider. So I've stubbed those
out. I also skipped the JWS payloads and just used a secret that's a
string. The JSON::JWT code seems smart enough to work out what to do
with all of this.

The important bit is that we're following the backchannel logout spec by
checking signatures for the logout token in the same way we're doing for
the ID token.

### DIFF
--- a/app/lib/oidc_client.rb
+++ b/app/lib/oidc_client.rb
@@ -3,6 +3,7 @@ require "openid_connect"
 
 class OidcClient
   class OAuthFailure < RuntimeError; end
+  class BackchannelLogoutFailure < RuntimeError; end
 
   attr_reader :client_id,
               :provider_uri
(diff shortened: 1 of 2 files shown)


alex9smith:
This looks good to me. I'll approve once you've either merged the other PR or changed the base of this one to that branch so I can see CI is happy with the unit tests, but I'm not expecting any issues

alex9smith (approved):

alex9smith:
Oh oops. Didn't mean to merge this!

PR #408, opened 2022-05-19
Remove cookies and feedback consent attributes

alex9smith:
We've removed the first time sign in page which asked users to consent
to cookies and sending them emails for feedback [[1]] and the form to
see and manage these attributes in the account home area [[2]].

Now we can remove them from the API and drop the database columns so
we're not hoarding personal data.

These two attributes are the only two `local` attributes, and are used
throughout for testing the logic for local vs. cached attributes.
Instead of removing all that logic as well, this commit sets up a new
dummy attribute `local_attribute` that can be used for testing while
keeping the logic in place. A later PR will clean up the different types
of attributes.

[1]: https://github.com/alphagov/frontend/pull/3234
[2]: https://github.com/alphagov/frontend/pull/3225

---

*DO NOT MERGE* until https://github.com/alphagov/frontend/pull/3225 and
the corresponding account management app PR have been merged. 

---
[Trello]()

### DIFF
--- a/app/controllers/internal/authentication_controller.rb
+++ b/app/controllers/internal/authentication_controller.rb
@@ -34,8 +34,6 @@ class Internal::AuthenticationController < InternalController
     render json: {
       govuk_account_session: govuk_account_session.serialise,
       redirect_path: redirect_path,
-      cookie_consent: govuk_account_session.user.cookie_consent,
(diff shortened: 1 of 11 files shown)


alex9smith:
I'm going to have to break this down further in order to get the pact tests passing.

PR #412, opened 2022-05-23
GUA-22 - Backchannel logout endpoint

huwd:
[Jira - GUA-22](https://govukverify.atlassian.net/browse/GUA-22)

Add a Backchannel logout endpoint which:
- Recieves a request with a logout_token as an signed JWT.
  - Responds with 400 if the JWT is invalid, or the signature cannot be validated
  - Responds 200 and creates a LogoutNotice if the JWT is valid.

A logout Notice is a Redis persisted `sub`, it is intended to sit there and await a visit from a user attempting to authorise with that sub. Upon success it will trigger a logout and redirect.

This PR only impliments the backchannel logout endpoint and expiry notice creation. Two follow up PRs will shortly be needed:

1. To check authorised internal requests and queries to the personalisation endpoint with a valid session cookie log out a user and redirect. This will need to _not_ apply to any login / create endpoints.
2. To invalidate / clear Redis upon a sucessful login or create attempt. We may also want to consider changes to the session duration on GOV.UK and if Redis based logout notices should be set to expire at a similar time.

Both of these will require a bit of 🧠 to consider a multi-device scenario, so I'm leaving these to a follow up.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -574,7 +574,7 @@ This endpoint requires the `update_protected_attributes` scope.
 #### Request parameters
 
 - `subject_identifier`
-  - the subject identifier of the user to delete
+  - the subject identifier of the user to update
 
 #### JSON request parameters
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -13,7 +13,7 @@ management. This API is not for other government services.
   - [`POST /api/oauth2/callback`](#post-apioauth2callback)
   - [`GET /api/oauth2/end-session`](#get-apioauth2end-session)
   - [`GET /api/user`](#get-apiuser)
-  - [`GET /user/match-by-email`](#get-apiusermatch-by-email)
+  - [`GET /user/match-by-email`](#get-usermatch-by-email)
   - [`GET /api/attributes`](#get-apiattributes)
   - [`PATCH /api/attributes`](#patch-apiattributes)
   - [`GET /api/email-subscriptions/:subscription_name`](#get-apiemail-subscriptionssubscription_name)
@@ -725,7 +725,7 @@ A public API endpoint that Implements OpenID Connect Back-Channel Logout 1.0 - d
 
 #### Note
 
-This endpoint is intended for calls to be made exclusively from the Digital Identity OP to the GOV.UK relaying party. As such it will not appear in GDS API Adaptors as it is not intended to be called from another Ruby compatible
+This endpoint will only ever receive calls from the Oidc OP (Digital Identity), as an OIDC event. Because of that, there is not an GDS API Adaptor for this endpoint, as it will not be called from another Ruby application.
 
 ## API errors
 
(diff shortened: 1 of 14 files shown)


docs/api.md
```diff
@@ -709,6 +710,23 @@ curl -H "GOVUK-Account-Session={{account_session_header}}" \
...
+  - A signed JWT
+
+#### Response codes
+
+- 400 if the Logout Token JWT is invalid or signature cannot be verified
+- 200 otherwise
+
+#### Note
+
+This endpoint is intended for calls to be made exclusively from the Digital Identity OP to the GOV.UK relaying party. As such it will not appear in GDS API Adaptors as it is not intended to be called from another Ruby compatible
```
alex9smith, line 728: I'm not sure what you mean by "not intended to be called from another Ruby compatible"

spec/requests/back_channel_logout_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      it "returns 400" do
+        post backchannel_logout_path, params: params
+        expect(response).to have_http_status(:bad_request)
+      end
+    end
+
+    context "with a valid logout token" do
+      it "records a session expiry notice" do
+        post backchannel_logout_path, params: params
+      end
```
alex9smith, line 41: I think you're missing the `expect` statement here

alex9smith (commented):
If we put this live now and give the URL to Auth, we'll start collecting `LogoutTokens` and not doing anything with them. That's fine, but we're coming up to the long bank holiday weekend and people have leave planned so I suspect we won't get round to finishing the logout part of this work until after that. Once we implement it, we'll suddenly log out every user we recieved a notice for over the last two weeks which doesn't sound ideal to me. 

What do you think about setting a short TTL on the `LogoutTokens` for now (1 day?) so we can test it, and then we can remove that once we implement the logout part?

spec/requests/back_channel_logout_spec.rb
huwd: quite right! woops

alex9smith (dismissed):

huwd:
Ok new plan, I got the invalidation rules done.
These are new in this PR and are present in the last 4 commits.

I think it makes sense to ship these together becuase then the notice will be added / removed by sensible rules.

The last PR will be to actually make the notice issue a logout - which I'll do now but merge after we've done a bit of testing as that's the first time we should have a user impact here.

I also added the Jira ticket to the commits becuase... I wanna see Jira join everything up nicely 😂

huwd:
Ace will merge this tomorrow (don't fancy the alerts if this breaks something now 😂)

The will rebase and take this off draft
https://github.com/alphagov/account-api/pull/423

If you want to get a head start over there [b7a4e44](https://github.com/alphagov/account-api/pull/423/commits/b7a4e4464d9a2f90e60c24575be1197128207c49) and [c872a54](https://github.com/alphagov/account-api/pull/423/commits/c872a5497e30b21471f019fa4f85fd2ce695d8e9) should be the only new bits!

alex9smith (approved):
Those 4 new commits look good to me as well 🚀

PR #418, opened 2022-05-25
Remove cookies and feedback consent attributes

alex9smith:
We've removed the first time sign in page which asked users to consent
to cookies and sending them emails for feedback [[1]] and the form to
see and manage these attributes in the account home area [[2]].

[[3]] Added a new local attribute that's used instead of these for
testing purposes.

Now we can remove them from the API.

[1]: https://github.com/alphagov/frontend/pull/3234
[2]: https://github.com/alphagov/frontend/pull/3225
[3]: https://github.com/alphagov/account-api/pull/410

---

*DO NOT MERGE* until https://github.com/alphagov/frontend/pull/3225 and
the corresponding account management app PR have been merged. 

---
[JIRA](https://govukverify.atlassian.net/jira/software/c/projects/GUA/boards/195?modal=detail&selectedIssue=GUA-11)


config/content_items.yml
```diff
@@ -34,10 +34,6 @@ special_routes:
     base_path: "/account/home"
     title: "Account home page"
     rendering_app: "frontend"
-  - content_id: "19c4cef8-d0ad-483b-93ea-e07f7ed4734a"
-    base_path: "/account/cookies-and-feedback"
```
huwd, line 38: Does this need to go at the same time?
https://github.com/alphagov/frontend/blob/6ad2e28c2dfa11357d2fe2b6b37a06406e011b24/test/functional/account_cookies_and_feedback_controller_test.rb

huwd: Also 🤔 about the special route publication here... we may need to recall these and publish a redirect?

alex9smith: Those are removed in https://github.com/alphagov/frontend/pull/3225 

Not sure about the redirect. In theory we've removed all places that link to this page, but we could publish a redirect just to be safe.

PR #421, opened 2022-05-25
Allow multiple Tombstone records for the same sub

alex9smith:
When this app was originally created it was *the* canonical place to
store account data. The Tombstone records were to allow us to count
deleted accounts and enforce uniqueness.

Now however, that's not the case. The source for account data is Auth's
OICD provider and Tombstones are less helpful to us now this app is
only part of a relying party's account storage.

Accounts on GOV.UK are only used for email notifications, so we plan to
use email alert API's logic to delete accounts that have no subscriptions.

This means it will be possible for someone to create an account, sign up
for notifications, delete that subscription and then have that account
deleted automatically. We need to make sure that if that person comes
back to GOV.UK and signs in, we can create a new `OidcUser` for them
that behaves exactly as expected.

In order for that to happen, we have to be able to create multiple
Tombstones for the same sub.

### DIFF
--- a/db/migrate/20220525152551_remove_unique_sub_index_from_tombstones.rb
+++ b/db/migrate/20220525152551_remove_unique_sub_index_from_tombstones.rb
@@ -1,0 +1,6 @@
+class RemoveUniqueSubIndexFromTombstones < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :tombstones, column: :sub
+    add_index :tombstones, :sub, if_not_exists: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_05_20_081234) do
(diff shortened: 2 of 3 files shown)


db/migrate/20220525152551_remove_unique_sub_index_from_tombstones.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveUniqueSubIndexFromTombstones < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :tombstones, column: :sub
+    add_index :tombstones, :sub, if_not_exists: true
```
huwd, line 4: huh, cool!
So that's a projection from any potential clash during migration?

https://blog.saeloun.com/2020/05/11/if-not-exists-for-index.html

Not come across that before

huwd (approved):

db/migrate/20220525152551_remove_unique_sub_index_from_tombstones.rb
alex9smith: Yeah exactly - we've got a lot of migrations going around at the moment so I thought it was safest

PR #422, opened 2022-05-25
Delete old Tombstone records

alex9smith:
In [[1]] we allowed there to be multiple Tombstones with the same sub
as this app is no longer the canonical place for account creation and
deletion data.

Since we're no longer checking for duplicates, we can also start
deleting old records as they're now only really useful for debugging.

We went with 30 days as that matches the time needed for an account to
be marked 'inactive' by email alert API.

We do lose the ability to count the number of accounts that has ever
existed by adding this worker but, as mentioned before, this app isn't
the right place to count this anyway so it's not a real loss.

[1]: https://github.com/alphagov/account-api/pull/421

### DIFF
--- a/app/models/tombstone.rb
+++ b/app/models/tombstone.rb
@@ -1,3 +1,6 @@
 class Tombstone < ApplicationRecord
   validates :sub, presence: true
+
+  EXPIRATION_AGE = 30.days
+  scope :expired, -> { where("created_at < ?", EXPIRATION_AGE.ago) }
 end
--- a/app/workers/expired_tombstone_worker.rb
+++ b/app/workers/expired_tombstone_worker.rb
@@ -1,0 +1,5 @@
+class ExpiredTombstoneWorker < ApplicationWorker
+  def perform
+    Tombstone.expired.delete_all
+  end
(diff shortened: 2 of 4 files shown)


alex9smith:
🤦🏻

KludgeKML:
Just a note: the exercises you folks gave Find and View to prepare us for taking over Accounts API have us looking into Tombstones as the way to tell whether an account ever existed. If you're changing that by deleting old tombstones you're going to need to update / rethink those exercises and warn us.

huwd:
Great point @KludgeKML (this will change that) we're making some fairly sweeping changes right now as part of our final withdrawal.
I've largely been getting on with those - but will need to come tell your team what we've done (the good new is there's now a lot less to worry about).

Wasn't sure if you'd all done the exercises by now, or if they're still ongoing?

KludgeKML:
Not sure about explore, but accessibility (who have the account-api ownership now the team responsibilities have been split) have mostly finished it (I think one person has a handful of exercises to do, but you can assume as a team we've basically completed everything).

huwd (approved):

PR #423, opened 2022-05-25
GUA-22 enforce logout notice

huwd:
Dependent on:
https://github.com/alphagov/account-api/pull/412

Enforces Logging users out where a LogoutNotice exists.
Applies to all Personalisation Routes (enforced at parent personalisation controller level)
Also applies to Internal Routes that use the AuthenticatedApiConcern

Must not apply to any routes that indicate an action that would supercede or invalidate a logout notice (for instance any logging in or creation action).

### DIFF
--- a/app/controllers/concerns/authenticated_api_concern.rb
+++ b/app/controllers/concerns/authenticated_api_concern.rb
@@ -10,7 +10,11 @@ module AuthenticatedApiConcern
         session_secret: Rails.application.secrets.session_secret,
       )
 
-      head :unauthorized unless @govuk_account_session
+      if @govuk_account_session
+        head :unauthorized if LogoutNotice.find(@govuk_account_session.user_id)
+      else
+        head :unauthorized
+      end
     end
 
     rescue_from AccountSession::ReauthenticateUserError do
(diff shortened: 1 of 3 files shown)


alex9smith:
Could you add a test similar to the check email one to any of the controllers that use the `AuthenticatedApiConcern`? I think we should play it safe just in case.

Other than that this looks good!

alex9smith:
Could you add a test similar to the check email one to any of the controllers that use the `AuthenticatedApiConcern`? I think we should play it safe just in case.

Other than that this looks good!

KludgeKML:
@alex9smith  / @huwd - is this PR still relevant?

alex9smith:
This looks like it's part of [implementing back-channel logout](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/log-your-users-out/#responding-to-logout-notifications-from-gov-uk-one-login), but based on my comment there we didn't finish that work. 

I'd say the work is still relevant and possibly worth doing, but the best way of doing that is probably to close this PR and for your team to tackle it as part of your backlog when it makes sense to do so.

KludgeKML:
Closing, work and reference captured here: https://trello.com/c/DJ3IB8kj/2285-implement-back-channel-logout-in-account-api

PR #573, opened 2023-01-06
Stop using OIDC autodiscovery.

sengi:
We don't need it; we control the only identity provider and the use of discovery led to the use of an entire memcache instance just to cache the response.

Let's get rid of this complexity because it's considerable extra work for Replatforming.


1pretz1 (approved):
Looks reasonable to me 👍

PR #629, opened 2023-04-18
Bump ruby version to 3.2.x

KludgeKML:
Trello: https://trello.com/c/ZwrQFAr0/1979-ruby-32-update-account-api

In this PR we're actually using 3.2.2 for local development, but freeing up the app to use 3.2.<latest> automatically when deployed, so that we don't have to do an update to benefit from patch versions, so this PR will need to be checked by replatforming.


1pretz1 (approved):

sengi (approved):
Glad to see it works with different patch versions 🎉 Thanks for pioneering this!

PR #689, opened 2023-08-08
Add Release workflow to version deployments

theseanything:
This adds a release workflow that creates a new GitHub Release (and git
tag) with an incremental version number. This version tag is then used
to tag images and referenced in the rest of the deployment system. This
makes it easier to compare references (vs commit SHAs). The version tag
is calculated as the last version incremented by the number of merge
commits until the reference commit.


theseanything:
Test bulk - recreating

PR #742, opened 2023-10-31
Delete subscriptions on manually deleting account

1pretz1:
Email subscriptions get ended when an account is deleted via the API [1], however the manual rake task doesn't include this step. If the subscriber doesn't exist then we'll just return `nil` and continue with the account deletion.

[1]: https://github.com/alphagov/account-api/blob/39af89177ed2ed6380d30f3469e6ab7c86348de2/app/controllers/internal/oidc_users_controller.rb#L31


KludgeKML (approved):
LGTM

PR #869, opened 2024-05-02
Set `.ruby-version` to `3.2.4`

pezholio:
This is similar to https://github.com/alphagov/whitehall/pull/9017 - unfortunately `rbenv` doesn't support minor versions


sengi (approved):
Thanks!

PR #919, opened 2024-07-08
Rename parameter variable

KludgeKML:
- in tests, this worked fine. But in practise, there was some odd variable shadowing going on that led to the client logout_token method being called with a nil.
 
Some details of thse issue:
-  https://gds.slack.com/archives/CADGP8PB4/p1720453623685349 / 
- https://stackoverflow.com/questions/15789540/methods-local-variable-with-same-name-as-another-method

(Basic problem: ruby instantiates local variables instantly and they override class methods, so by declaring a variable logout_token, even on the left hand side, we instantly create that token as nil. Then when it's passed to the variable on the right hand side it has that nil value rather than the one from the method).

### DIFF
--- a/app/controllers/oidc_events_controller.rb
+++ b/app/controllers/oidc_events_controller.rb
@@ -1,6 +1,6 @@
 class OidcEventsController < ApplicationController
   def backchannel_logout
-    logout_token = oidc_client.logout_token(logout_token)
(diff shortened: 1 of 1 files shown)


beccapearce (approved):
OOH nicely worked out! 🧠 Very niche ruby!

